PR #1, opened 2017-06-27
initial setup

nslater:


requirements.txt
```diff
@@ -0,0 +1,4 @@
+crate
+crate-docs-theme
+sphinx-autobuild
+Sphinx
```
chaudum: please specify a specific version

requirements.txt
```diff
@@ -0,0 +1,4 @@
+crate
```
chaudum: crate package is not necessary

PR #17, opened 2017-08-02
import best practices from the reference

nslater:
see https://trello.com/c/y9C6TSES/30


nslater:
see also crate/crate#6062

PR #21, opened 2017-08-21
cross reference the crate reference

nslater:


nslater:
see also crate/crate#6174

PR #26, opened 2017-09-21
add documentation for using Telegraf with CrateDB

tomsvogel:


nslater:
@tomsvogel please review my pr against your branch: #28

once my pr is merged, I will have additional comments for you on these changes!

thank you!! :)

docs/integration/telegraf.txt
```diff
@@ -0,0 +1,66 @@
+===============================
+Using Telegraf with CrateDB
+===============================
+
+`Telegraf <https://github.com/influxdata/telegraf>`_ is a plugin-driven server agent for collecting & reporting metrics.
```
nslater: okay so for this intro section I think we need a bit more info

can you add a para saying why you might want to use CrateDB with Telegraf (use case, etc) and what this particular doc is going to show you how to do and for what reason

docs/integration/telegraf.txt
```diff
@@ -0,0 +1,66 @@
...
+  # See http://godoc.org/github.com/lib/pq#hdr-Connection_String_Parameters
+  url = "postgres://crate@localhost/doc?sslmode=disable"
+  # Timeout for all CrateDB queries.
+  timeout = "5s"
+  # Name of the table to store metrics in.
+  table = "metrics"
+  # If true, and the metrics table does not exist, create it automatically.
+  table_create = true
+
+Run
```
nslater: after this section, it would be good if there was a section telling the reader what they can expect. or how to test that it worked. or something along those lines

docs/integration/telegraf.txt
tomsvogel: Is this enough? or feel free to add something.

docs/integration/telegraf.txt
```diff
@@ -0,0 +1,69 @@
+===============================
+Using Telegraf with CrateDB
+===============================
+
+`Telegraf <https://github.com/influxdata/telegraf>`_ is a plugin-driven server agent for collecting & reporting metrics.
+Unsing Telegraf in combination with CrateDB allows you collecting a huge amount of data and query them with SQL in real-time.
```
nslater: can you make it more specific why crate is a good fit here. is telegraf not good at doing this by itself? why would you want to add crate to the mix?

docs/integration/telegraf.txt
nslater: this is fin

docs/integration/telegraf.txt
joemoe: @nslater here are some additional reasons:
- Prometheus' internal storage has limited scalability.
- Users want to use the data outside of Prometheus to visualize it or run application code.
- They want to use CrateDB as long time storage.

Hope that's sufficient.

joemoe: @nslater Telegraf is a middleware connects inputs with outputs, so you may want to collect data with Kafka, Prometheus and what so ever and move them to what ever output like CrateDB. So you don't need to use CrateDB when using Telegraf. But you might have some input sources that you want to connect with CrateDB. Telegraf helps you get the data there. See this list https://github.com/influxdata/telegraf#input-plugins
You may also want to process or aggregate data before it is ingested into CrateDB.

nslater:
@tomsvogel please review my edits pr! #34

docs/integration/telegraf.txt
nslater: thanks @joemoe

PR #29, opened 2017-09-27
add docs for insert performance

nslater:
this pr:

- adds a performance/inserts section to the guide
- will collect four additional prs, each one adding a new page to this new section

once all four prs are merged, this pr can be merged to master


docs/performance/inserts/tuning.txt
```diff
@@ -0,0 +1,172 @@
...
+
+
+Here, the most interesting metric is ``%iowait``. This tells you how
+time the CPU is waiting for disk I/O. You can expect better insert performance
+from CrateDB relative to how low this value is.
+
+CPUs
+----
+
+If your queries are complex, the additional CPU overhead during the
```
mfussenegger: This has more to do with the table schema / number of columns. If fulltext/analyzers are in use, those can be CPU heavy.

docs/performance/inserts/tuning.txt
```diff
@@ -0,0 +1,172 @@
...
+:ref:`analyzing phase <parsing_analyzing>` of insert query processing might
+result in a CPU performance bottleneck.
+
+If this becomes the case, and simplifying your queries isn't an option,
```
mfussenegger: Same here, it's not so much about the query, but about the schema. So simplifying would in this case mean to change the table schema to use different analyzers, or even use `INDEX OFF`

docs/performance/inserts/tuning.txt
```diff
@@ -0,0 +1,172 @@
...
+------
+
+Replicas
+........
+
+Biggest influence: replicas (esp 0 vs . 1+)
+
+(Because primary write first and only after it writes the replicas)
+
+TODO: need details. what are we recommending here?
```
mfussenegger: 0 should almost never be used (except maybe temporary for initial imports) 
At least 1 for a production 3 cluster setup.

More depends on the expected availability. (And going from 1 to 2 isn't as costly as from 0 to 1 in terms of insert duration, load increases linearly)

docs/performance/inserts/tuning.txt
```diff
@@ -0,0 +1,172 @@
...
+
+Shards and Partitioning
+.......................
+
+Number of shards / partitioning
+
+TODO: need details. what are we recommending here?
```
mfussenegger: Don't we have a sharding guide already that we could link here?

docs/performance/inserts/tuning.txt
```diff
@@ -0,0 +1,172 @@
...
+
+Table Schema
+............
+
+(Analyzer, INDEX OFF, Primary Key)
+
+TODO: need details. what are we recommending here?
```
mfussenegger: Setting `INDEX OFF` is a good way to increase index performance, but those columns won't be useable in the WHERE clause. So this is really use-case dependend.

For primary keys: If there is no "natural" primary key due to the use case, it's better to avoid using an explicit primary key, but instead rely on the internal `_id`

docs/performance/inserts/tuning.txt
```diff
@@ -0,0 +1,172 @@
...
+
+Translog
+........
+
+https://crate.io/docs/crate/reference/en/latest/sql/reference/create_table.html#translog-durability
+
+TODO: need details. what are we recommending here?
```
mfussenegger: Again, depends on the use-case/requirements. `ASYNC` leads to more performance, but bigger chance for data loss in case of disk failures. `REQUEST` is the default due to that.

docs/performance/inserts/architecture.txt
```diff
@@ -0,0 +1,263 @@
...
+
+Execution
+---------
+
+The planned inserts are then handed off to the *executor*, which executes them
+across the cluster.
+
+This process is managed by the node responsible for handling the client
+request, but the operations needed to complete the inserts can be handed off to
+any `data node`_ in the cluster.
```
mfussenegger: hm.. I think here it would be nice to have some kind of illustration. E.g

```

    +--------+                  +---------+
    | Client |----- insert -->  | Handler |
    +--------+     (values)     +---------+
                                  |
                         ---------+----
                       _/              \_
                      /                  \
                     /                    \
                    insert               insert
                 +---------+          +---------+
                 |  Node   |          |  Node   |
                 | Shard 1 |          | Shard 2 |
                 +---------+          +---------+


    +--------+                 +---------+
    | Client |----- insert --> | Handler |
    +--------+    (subquery    +---------+
                   from table)   |
                         +--------++
                       _/           \_
                      /               \
                     /                 \
                    select           select
                 +---------+       +---------+
                 |  Node   |       |  Node   |
                 | Shard 1 |       | Shard 2 |
                 +---------+       +---------+
                    |  \            /  |
                    |   \          /   |
                    |    \________/    |
                    |    /        \    |
                    |   /          \   |
                    |  /            \  |
                    insert           insert
                 +---------+       +---------+
                 |  Node   |       |  Node   |
                 | Shard 1 |       | Shard 2 |
                 +---------+       +---------+
```

docs/performance/inserts/architecture.txt
```diff
@@ -0,0 +1,263 @@
...
+
+How the executor proceeds depends on the type of inserts being executed.
+
+Executing Single Inserts
```
mfussenegger: I wonder if it would make sense to restructure this as well.
The execution logic is always the same: 

On "sender" (handler in case of values/unnest; data node in case of insert from subquery)
`data -> group by shards -> create missing partitions -> invoke index action per shard`

where invoke index action does the steps that are already described below
(check number of replicas, do the actual write, fire off replica requests)

The only difference between bulk/single insert is that the response created is different. (single inserts produces a failure message, in bulk actions we just set the rowcount to -2 for the position)

docs/performance/inserts/architecture.txt
```diff
@@ -0,0 +1,263 @@
...
+- If set to some value ``x``, the insert will only proceed once the primary
+  shard plus ``x-1`` replica shards are available
+
+- If set to ``1``, the insert will proceed if the primary shard is available
+
+If the required number of replicas is not available, and do not become
+available during a brief waiting period, the operation fails and an error is
+returned back to the user.
+
+TODO IS THE WAITING PERIOD CONFIGURABLE?
```
seut: no it isn't

docs/performance/inserts/architecture.txt
```diff
@@ -0,0 +1,263 @@
...
+expressions, as listed under the previous header, with one exception.
+
+Instead of reporting a single number corresponding to the number of successful
+inserts, the previously created bulk insert response object is populated with
+information about every insert attempted.
+
+The `bulk insert response`_ object is returned to the client, with a
+``rowcount`` for every row in the bulk insert request:
+
+- A ``rowcount`` of ``1`` indicates a successful insert
```
seut: don't get this, the rowCount will express the successfully inserted documents, so 0 or greater

docs/performance/inserts/architecture.txt
```diff
@@ -0,0 +1,263 @@
...
+
+- A ``rowcount`` of ``2`` indicates a failed insert, and may be
```
seut: `-2` (not 2) expresses an error of this bulk operation, no record related to this bulk was inserted

docs/performance/inserts/architecture.txt
```diff
@@ -0,0 +1,263 @@
...
+Planning Inserts With Multiple Value Expressions
+................................................
+
+For each row to be inserted, the planner performs the following actions:
+
+1. If the partition doesn't exist, create a new partition
+
+2. Calculate the shard ID (``ShardId``) using the `routing column`_
+
+3. Create a single insert request for the shard ID
```
seut: rows are grouped per shard, so added a shard request (request creation happens lazy underneath if no one exists for that shard yet to be accurate but afaik thats irrelevant here)

docs/performance/inserts/architecture.txt
seut: yeah that would be nice. I think inserts-by-query are in general missing at this document yet

docs/performance/inserts/parallel.txt
```diff
@@ -0,0 +1,131 @@
+.. _parallel_inserts:
+
+================
+Parallel Inserts
+================
+
+Because CrateDB is a clustered data with every node capable of handling client
+requests, parallelizing (i.e. doing many at once) your inserts is one good way
```
mxm: What would be another way?

docs/performance/inserts/parallel.txt
```diff
@@ -0,0 +1,131 @@
...
+to increase insert performance.
+
+There are a few ways you can do this.
```
mxm: In the following, a few ways how you can do this

docs/performance/inserts/architecture.txt
```diff
@@ -0,0 +1,263 @@
...
+- Do the specified table columns exist?
+
+   - If not, does the table have a `dynamic column policy`_?
+
+- Do the data types of the insert values match the data types of the existing
+  columns?
+
+- Are all ``PRIMARY KEY`` columns and ``NOT NULL`` column specified?
+
+- Are the values for ``NOT NULL`` columns non null?
```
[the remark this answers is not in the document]
nslater: addressed

docs/performance/inserts/parallel.txt
```diff
@@ -0,0 +1,131 @@
...
+resources on that node can easily become a bottleneck.
+
+Switch Between Multiple Nodes
+-----------------------------
+
+Configure your client application with a list of nodes in your cluster, and
+select a random node each time a request is made.
+
+The CrateDB `Python client`_, `JDBC client`_, and `PHP PDO client`_ support some
+sort of multiple node configuration.
```
mxm: "some sort" sounds vague. would leave these words out.

docs/performance/inserts/parallel.txt
```diff
@@ -0,0 +1,131 @@
...
+
+You can implement your own too.
```
mxm: Again, this sounds vague and not really helpful. You can always implement it yourself :)

docs/performance/inserts/parallel.txt
```diff
@@ -0,0 +1,131 @@
...
+  If you have a small number of clients, they will contact a small number of
+  nodes in your cluster.
+
+  DNS cache time (the *TTL* value) can be reduced, but this results in
+  additional DNS overhead, and doesn't help much if you have a large number of
+  nodes and small number of clients who are regularly stuck making requests to
+  a single node.
+
+  Ideally, clients should be rapidly switching beween CrateDB nodes, to make
+  best use of the cluster.
```
mxm: Not sure if this DNS method is a good way of load balancing. You even say that the previous method is better here.

Maybe re-arrange the sections, first single node, then DNS, then multiple nodes, then connection pooling, then load balancer?

docs/performance/inserts/parallel.txt
```diff
@@ -0,0 +1,131 @@
...
+
+Use a Load Balancer
+-------------------
+
+Stick your CrateDB cluster in front of a load balancer.
+
+With something like `HAProxy`_, you can configure a round robin load balancing
```
mxm: "something" - you like this word :) I'd say "With a load balancer proxy like HaProxy"...

docs/performance/inserts/parallel.txt
```diff
@@ -0,0 +1,131 @@
...
+This technique comes with the added benefit that most load balancers (like
+HAProxy) will monitor the cluster and provide health checks, analytics
+dashboards, and so on.
+
+Client Connection Pooling
+-------------------------
+
+If your client supports it, connection pooling can be a good way to both load
+balance requests as well as make use of connection reuse. This may offer
+performance benefits over a fully featured load balancer.
```
mxm: How's this section different from "Switch Between Multiple Nodes"? That's what the connection pools does as well.

docs/performance/inserts/parallel.txt
```diff
@@ -0,0 +1,131 @@
...
+Follow the instructions in our :ref:`testing_inserts_performance` guide.
+
+To test parallel inserts, you should:
+
+1. Configure the setup you would like to test
+
+2. Run a number of different tests against that setup, using different
+   ``--concurrency`` settings
+
+3. Plot your results on a graph to see the performance curve of your setup
```
mxm: Not sure if plotting is necessary. We could just say "evaluate the throughput results".

docs/performance/inserts/parallel.txt
```diff
@@ -0,0 +1,131 @@
...
+   (i.e. how your setup responds to different loads)
+
+4. Start over
```
mxm: 😆 Maybe more like "Try out different settings and re-run the experiment.

docs/performance/inserts/parallel.txt
```diff
@@ -0,0 +1,131 @@
...
+
+At the end of this process, you will have a number of performance curve graphs
+for the different setups you are interested in testing.
```
mxm: This sounds vague. 

"At the end of this process, you will have a better understanding of the throughput of your cluster with different concurrency levels".

docs/performance/inserts/architecture.txt
nslater: hmm

that's not my understanding of this doc:

https://crate.io/docs/crate/reference/en/latest/protocols/http.html#bulk-errors

this seems to indicate a separate response for every insert. indeed this appears to be the advantage to using this method over the previous method (that just tells you the total inserted rows with no details about errors)

am I misunderstanding something

docs/performance/inserts/architecture.txt
nslater: again, going from this:

https://crate.io/docs/crate/reference/en/latest/protocols/http.html#bulk-errors

I thought that this object is returning fine-grain results for every single row to be inserted. so you either get "1" row inserted or an error

docs/performance/inserts/parallel.txt
nslater: well, we're half way through a second in the docs called "inserts performance" so I am presuming that from context it's clear that this is one of multiple approaches being suggested

thoughts?

docs/performance/inserts/parallel.txt
nslater: done

docs/performance/inserts/testing.txt
```diff
@@ -0,0 +1,96 @@
...
+    cr> CREATE TABLE my_table (
+    ...   id integer,
+    ...   name string
+    ... );
+    CREATE OK, 1 row affected  (... sec)
+
+This is a simple example. You should create a table that mirrors the sort of
+data you plan to use in production.
+
+Next, generate some fake data:
```
mfussenegger: Ideally they'd use actual data that they plan to store - this should be as a fallback used with caution because the data will very likely be different

docs/performance/inserts/testing.txt
```diff
@@ -0,0 +1,96 @@
+.. _testing_inserts_performance:
+
+===========================
+Testing Inserts Performance
+===========================
+
+Every setup is different. The best way to tune your particular setup is to do
+performance testing.
```
mfussenegger: I'm wondering if it would make sense to point out that it's reasonable to test against a single node first, since inserts scale pretty much linearly.

docs/performance/inserts/architecture.txt
seut: hm ok right somehow. I was a referring to a bit special case where one is mixing bulk request with multiple value inserts which we do support. like using following insert statement with bulk arguments:
```
insert into t1 (id, name) values (?, ?), (?, ?)
```

in this case every bulk result will return `2` as a rowcount if both inserts were successful. 
but I'm not sure if we should pointing out this case, maybe we can just rephrase it to be a bit more generic.

docs/performance/inserts/parallel.txt
```diff
@@ -0,0 +1,130 @@
+.. _parallel_inserts:
+
+================
+Parallel Inserts
+================
+
+Because CrateDB is a distributed database with every node capable of handling
+client requests, parallelizing your inserts (i.e. doing many at once) is a
```
mfussenegger: Maybe we could also explain why that's a good approach?
E.g. mention that users have to make sure that the client isn't the bottle neck. E.g. that if there is a latency of 5ms between client and server, a client doing sequential inserts couldn't reach more than 200 inserts/s even if the insert on the server side required 0ms processing time. Concurrent or parallel inserts can help there (there is a difference between concurreny and parallelism, though it's often used interchangeable..)

docs/performance/inserts/architecture.txt
seut: yes right, its for every bulk result. but its `-2` not `2` ;) sry if I wasn't clear enough.

docs/performance/inserts/architecture.txt
nslater: if you do a bulk insert with multiple values, what happens when there's an error? how are errors distinguished from correct row count success?

docs/performance/inserts/parallel.txt
```diff
@@ -0,0 +1,130 @@
...
+Use a Single Node
+-----------------
+
+Point your client application at a single node in your cluster, and send many
+concurrent requests.
+
+This is the least efficient way to do parallel inserts, because the
+resources on that node can easily become a bottleneck.
+
+DNS Round Robin
```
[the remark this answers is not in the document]
nslater: yes it would. can you tell me what order that is? hehe

docs/performance/inserts/parallel.txt
```diff
@@ -0,0 +1,130 @@
...
+    server cratedb1 192.0.2.1:4200 check
+    server cratedb2 192.0.2.2:4200 check
+    server cratedb3 192.0.2.3:4200 check
+
+With this setup, your client applications would all point directly at the load
+balancer. The load balancer then distributes requests across your cluster.
+
+This technique comes with the added benefit that most load balancers (like
+HAProxy) will monitor the cluster and provide health checks, analytics
+dashboards, and so on.
```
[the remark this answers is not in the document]
nslater: okay

does it make sense to mention proxies that can handle the postgre wire protocol? do we know of any that are any good?

I found this one:

https://github.com/CrunchyData/crunchy-proxy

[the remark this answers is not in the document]
nslater: are there no proxies that can do switch overs? i.e. detect dead nodes and remove them from the backend pool? I'm not v experienced here

docs/performance/inserts/parallel.txt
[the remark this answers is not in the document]
nslater: should we also cover discovery via dns and aws/azure?

https://crate.io/docs/crate/reference/en/latest/configuration/cluster.html#discovery-via-dns

are there any clients or client libraries that can build the node list this way?

docs/performance/inserts/parallel.txt
nslater: should we rename this section to "concurrent requests"? and mention the parallel difference. or the other way around?

docs/performance/inserts/testing.txt
nslater: done

docs/performance/inserts/testing.txt
nslater: done. added whole new section

docs/performance/inserts/parallel.txt
mfussenegger: I'd rename it to concurrent inserts - I wouldn't mention the parallel difference. But it might make sense to explain the latency aspect.

docs/performance/inserts/parallel.txt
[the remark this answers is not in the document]
mfussenegger: > can you tell me what order that is

I'd say client-round-robin | connection pooling -> DNS | Load balancer

I don't think there are currently clients which support dns or aws/azure discovery, so I wouldn't mention that.

docs/performance/inserts/parallel.txt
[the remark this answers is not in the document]
mfussenegger: > can you tell me what order that is

They do detect dead nodes and remove them, but afaik they can't recover from an error *happening*. If the client sens a request that is redirected to node just dying shortly after the redirect decision was made. But I'm not sure.

> does it make sense to mention proxies that can handle the postgre wire protocol? do we know of any that are any good?

I've no expierence with any of those. 

Hm - giving this some more thought. I wonder if the current structure should be changed here slightly as well. I see two points being made here: 

- Make sure the client isn't the bottle neck (Do concurrent inserts)
- Make sure a single node isn't the bottle neck (Spread the load)

Maybe we could represent that better in the structure and on each point list the content we have already as options to achieve that?

docs/performance/inserts/parallel.txt
nslater: done

docs/performance/inserts/parallel.txt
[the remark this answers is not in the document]
nslater: you're right! I'm creating a new doc in the performance section that is for load balancing. bcus this will affect other aspects of performance

docs/performance/inserts/parallel.txt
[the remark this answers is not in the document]
nslater: okay thanks!

docs/performance/inserts/architecture.txt
nslater: okay this is being addressed with the rewrite

docs/performance/inserts/architecture.txt
nslater: this is being addressed with the rewrite

docs/performance/inserts/architecture.txt
```diff
@@ -0,0 +1,263 @@
...
+responsible for handling the client request.
+
+How the planner proceeds depends on the type of insert being prepared.
+
+Planning Single Inserts
+.......................
+
+The planner performs the following actions:
+
+1. If the partition doesn't exist, create a new partition
```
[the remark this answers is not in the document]
nslater: okay this is being addressed with the rewrite

nslater:
considering how many structure changes we've made to these docs since opening the pr. once I've finished up addressing some of the todos in the files, and once I've merged the rewritten architecture doc back into the source tree, I will delete this pr and open a new one

PR #32, opened 2017-09-28
add page on bulk inserts

nslater:
for #29


docs/performance/inserts/bulk.txt
```diff
@@ -0,0 +1,147 @@
...
+
+Fortunately, there is a solution: bulk inserts.
+
+Switching to Bulk Inserts
+=========================
+
+If you can batch up your inserts, switching to *bulk inserts* will dramatically
+improve both ingestion performance, and overall cluster performance.
+
+Bulk inserts make use of the CrateDB HTTP endpoint `bulk operations`_ feature
```
mfussenegger: Some postgres based clients also provide functionality which let us create/queue up a bulk request internally. (There are still individual messages from the client, but these are lower overhead because they're binary and contain almost no headers and happen on an already established connection).

Furthermore, bulk insert can also be accomplished by using multiple values clauses (though it's not *exactly* the same, as using the bulk operations features)

docs/performance/inserts/bulk.txt
```diff
@@ -0,0 +1,147 @@
...
+to perform many inserts in a single operation.
+
+Bulk inserts still generate internal network traffic, and CrateDB still waits
+until all replicas have been updated before returning a response. But there is
+less internal network traffic. And in addition, the bulk query only needs to be
```
mfussenegger: maybe it's worth mentioning that the disk flush only happens once as well (if set to request, which is the default, see https://crate.io/docs/crate/reference/en/latest/sql/reference/create_table.html?highlight=flush#translog-durability)

docs/performance/inserts/bulk.txt
```diff
@@ -0,0 +1,147 @@
...
+performance testing.
+
+The easiest way to do this is to use the `cr8`_ tool.
+
+Let's say you have CrateDB running on ``localhost:4200``, and you want to test
+inserts against the ``my_table`` table. You can run cr8, like so:
+
+.. code-block:: sh
+
+    cr> time cr8 insert-fake-data \
```
mfussenegger: `insert-fake-data` wasn't built with benchmarking in mind. Depending on the schema it could be that the random-data generation is the bottle neck. A better way would be to generate a json file  use `insert-json` (insert-fake-data could be used to generate the data first, then use copy to to export and re-import via insert-json)

`insert-json` will also output average/min/max and percentile values, which is more useful then the simple time based average from `insert-fake-data`

docs/performance/inserts/bulk.txt
nslater: @mfussenegger can you provide the names of the clients that offer bulk requests? we don't have to list all of them. but the the noteworthy ones would be good

also, re using multiple values clauses. I tried this and couldn't get it to work. I also couldn't see anything in the docs. can you provide an example query?

mfussenegger: I guess the noteworthy one is JDBC (`addBatch` and `executeBatch`),  as it's the only one where we also have tests for it.

multiple values: `insert into t (x, y) values (1, 1), (2, 2), (3, 3)`

docs/performance/inserts/bulk.txt
```diff
@@ -93,6 +98,52 @@ Combine ``UNNEST`` with ``INSERT`` to perform bulk inserts:
...
+
+.. code-block:: psql
+
+    cr> INSERT INTO my_table (id, name)
+    ...      VALUES (1, 'Arthur'),
+    ...             (2, 'Trillian'),
+    ...             (2, 'Marvin');
+    INSERT OK, 3 rows affected  (... sec)
+
+This method of doing bulk inserts is just as performant as the ``UNNEST``
```
mfussenegger: With a large amount of values this will actually be slower because parsing the statement becomes more expensive.

docs/performance/inserts/bulk.txt
```diff
@@ -0,0 +1,258 @@
...
+    |    3 | Marvin   |
+    +------+----------+
+    SELECT 3 rows in set  (... sec)
+
+Combine ``UNNEST`` with ``INSERT`` to perform bulk inserts:
+
+.. code-block:: psql
+
+    cr> INSERT INTO my_table (id, name)
+    ...      VALUES (SELECT *
```
mfussenegger: `VALUES` needs to be removed here, it's just the select

docs/performance/inserts/bulk.txt
```diff
@@ -0,0 +1,258 @@
...
+If your client supports query string parameter substitution, you can use the
+``UNNEST`` method with static prepared statements.
+
+For example, using the CrateDB Python client, the following is possible:
+
+.. code-block:: python
+
+    client.execute("""
+      INSERT INTO my_table (id, name)
+           VALUES (SELECT * FROM UNNEST(?, ?))
```
mfussenegger: Without `VALUES`

PR #40, opened 2017-10-25
 add docs for insert performance

nslater:
replacement for https://github.com/crate/guide/pull/29


docs/performance/inserts/parallel.txt
```diff
@@ -0,0 +1,88 @@
...
+                // use row count
+            } else {
+                // handle insert failure
+            }
+        });
+    });
+```
+This way the inserts will be supplied to be executed asynchronously on the 
+``commmonPool`` (you can also provide your own ``Executor`` using the 
+appropriate ``supplyAsync`` signature).
```
mfussenegger: Maybe we should also add a note to clarify that this if possible, it would also greatly improve performance if the inserts were batched and cross-link to the other documents? (it would actually the first thing that should be done if possible - before trying to add parallelism)

docs/performance/inserts/parallel.txt
```diff
@@ -0,0 +1,88 @@
...
+Follow the instructions in our :ref:`testing_inserts_performance` guide.
+
+To test parallel inserts, you should:
+
+1. Configure the setup you would like to test
+
+2. Run a number of different tests against that setup, using different
+   ``--concurrency`` settings
+
+3. Evaluate your throughput results (perhaps by plotting your results on a
```
mfussenegger: btw if we wanted to further boost this content I think we could also add a concrete example walk-through here.

docs/performance/inserts/parallel.txt
andreidan: I agree. I think that should be part of the introduction though, saying that explicit parallelism should be explored in case batching writes doesn't already provide the desired performance or something along those lines.

docs/performance/inserts/parallel.txt
nslater: I think this might be better content for the blog, and cross link! I'll add a note

nslater:
@mfussenegger re this comment:

https://github.com/crate/guide/pull/40#discussion_r151086694

can you clarify what you mean please? do you mean to do parallel inserts with the bulk api endpoint? that would require another, different, code sample, correct? or do you mean prepared statements? or some other thing?

nslater:
@mfussenegger I have addressed your comment in your latest fixup. this pr is still not ready for a re-review just yet though

nslater:
@mfussenegger this is ready for review

this work was being held up by a work in progress doc that will eventually cover load balancing. but that work should have been done as a separate PR in the first place, so I have moved it off and created a new card to track it

there were a few minor todos in the inserts content that I have either fixed or moved off to be tracked as separate work

docs/performance/inserts/methods.txt
```diff
@@ -0,0 +1,283 @@
...
+ 1. Every insert is first applied to the primary shard
+
+ 2. After the primary shard has been updated, the insert is then individually
+    communicated in parallel to every configured replica shard
+
+ 3. CrateDB will not return a response until all replica shards have been
+    updated
+
+The overhead for each insert (query parsing, planning, and execution, which
+includes network connection setups, connection teardowns, and connection
```
mfussenegger: Not sure if we should include connection setups and teardown here. Node-to-node the connections are kept alive. Similar with HTTP which supports keep-alive. Postgres connection is also kept open until the client disconnects.

docs/performance/inserts/methods.txt
```diff
@@ -0,0 +1,283 @@
...
+
+You should see a dramatic improvement in performance over single inserts.
+
+Specifically, the advantages are:
+
+- Significantly less internal network traffic
+
+- The query only needs to be parsed, planned, and executed once
+
+- If `translog.durability`_ is set to ``REQUEST`` (the default), an insert
```
mfussenegger: It's once per shard that is being written to. Depending on the amount of rows being inserted and the number of shards you might end up writting to multiple shards.

docs/performance/inserts/methods.txt
```diff
@@ -0,0 +1,283 @@
...
+In addition to reducing parsing overhead, prepared statement execution requests
+use the binary protocol, contain almost no headers, and are executed over an
+already established connection.
+
+Typically, prepared statements are used :ref:`single insert statements
+<inserts_single_inserts>`.
+
+Prepared statements with single inserts will usually perform better than
+:ref:`standard querying <inserts_standard_querying>` with single inserts, but
+worse than standard querying with both the :ref:`UNNEST method<inserts_unnest>`
```
mfussenegger: It should be comparable to unnest/multi-value expression as we internally try to execute the whole batch as one, so you've the same benefits (1 flush per shard being written to for the whole batch).

docs/performance/inserts/tuning.txt
```diff
@@ -0,0 +1,141 @@
...
+is better than creating your own `surrogate primary key`_ (e.g. manually
+generating a UUID for each row) because there is one less column to index.
+
+Translog
+--------
+
+If `translog.durability`_ is set to ``REQUEST`` (the default), the
+:ref:`translog <durability>` gets flushed after every operation. Setting this
+to ``ASYNC`` will improve insert performance, but it also worsens durability.
+If a node is shut down before a translog has been synced, those opperations
```
mfussenegger: Maybe replace `shut down` with `crashes` - on a regular controlled shutdown there won't be data loss

docs/performance/inserts/tuning.txt
```diff
@@ -0,0 +1,141 @@
...
+will be lost.
+
+Refresh Interval
+----------------
+
+Data that has been written to a shard cannot be read back until the shard index
```
mfussenegger: With primary key lookups it is possible to retrieve written data immediately without explicit refresh. - Maybe re-phrase this slightly to take that into account?

docs/performance/inserts/methods.txt
nslater: changed to:

> The overhead for each insert (query parsing, planning, and execution, and
potentially network connection overhead too) starts to add up for very heavy
insert workloads.

docs/performance/inserts/methods.txt
nslater: changed to:

> If `translog.durability`_ is set to ``REQUEST`` (the default), an insert
  using ``UNNEST`` flushes the disk once for every shard written to.

docs/performance/inserts/methods.txt
nslater: changed to:

> and should be comparable to standard querying with both

docs/performance/inserts/tuning.txt
nslater: done

docs/performance/inserts/tuning.txt
nslater: changed to: 

> With the exception of primary key lookups, data that has been written to a
shard cannot be read back until the shard index has been refreshed.

nslater:
@mfussenegger thank you! I believe I have addressed all of your comments. this PR is ready for your re-review

docs/performance/inserts/methods.txt
```diff
@@ -0,0 +1,284 @@
...
+The advantages are the same as using the :ref:`UNNEST method<inserts_unnest>`:
+
+- Significantly less internal network traffic than executing each insert
+  statement individually
+
+- Even though you're executing multiple insert statements, the bulk query only
+  needs to be parsed, planned, and executed once
+
+- If `translog.durability`_ is set to ``REQUEST`` (the default), a bulk insert
+  only flushes the disk once
```
seut: also here it should state `... once for every shard written to..`

nslater: resolving with https://github.com/crate/guide/pull/108

PR #55, opened 2017-12-12
replace "_" in urls with "-"

nslater:


nslater:
@Autophagy okay I have expanded the scope of this quite a bit. can you re-check pls?

nslater:
@Autophagy pushed a fixup

PR #63, opened 2017-12-12
add intro text to performance section

nslater:


docs/performance/index.txt
```diff
@@ -2,6 +2,9 @@
 Performance Tuning
 ==================
 
+This section of the documentation shows tune your setup and get the best
```
andreidan: "shows how to tune ..."

PR #69, opened 2018-01-11
fix terminology

nslater:


nslater:
@Autophagy done, thanks! can you recheck?

PR #73, opened 2018-01-26
Document the cratedb docker healthcheck.

andreidan:


docs/deployment/containers/docker.txt
```diff
@@ -191,6 +191,13 @@ Here's how you would mount ``crate.yml``::
 Here, ``/srv/crate/config/crate.yml`` is an example path, and should be
 replaced with the path to your host machine ``crate.yml`` file.
 
+Healthcheck
+-----------
+
+The official `CrateDB Docker image`_ ships with a configured liveness
+`Docker healthcheck`_. The purpose of this check is to signal the situations
+when the CrateDB process might've crashed or hanged inside the container
```
matriv: `'ve` to `have`

docs/deployment/containers/docker.txt
```diff
@@ -191,6 +191,13 @@ Here's how you would mount ``crate.yml``::
 Here, ``/srv/crate/config/crate.yml`` is an example path, and should be
 replaced with the path to your host machine ``crate.yml`` file.
 
+Healthcheck
+-----------
+
+The official `CrateDB Docker image`_ ships with a configured liveness
+`Docker healthcheck`_. The purpose of this check is to signal the situations
```
matriv: `signal` -> `indicate`

andreidan: @nslater what do you think? I find `signal` to be expressive enough

docs/deployment/containers/docker.txt
```diff
@@ -191,6 +191,13 @@ Here's how you would mount ``crate.yml``::
 Here, ``/srv/crate/config/crate.yml`` is an example path, and should be
 replaced with the path to your host machine ``crate.yml`` file.
 
+Healthcheck
+-----------
+
+The official `CrateDB Docker image`_ ships with a configured liveness
```
nslater: drop "liveness"

docs/deployment/containers/docker.txt
nslater: This health check will flag that CrateDB process might...

nslater: also change it to "Docker health check"

docs/deployment/containers/docker.txt
andreidan: I don't agree with this as there can be multiple types of checks (ie. readiness - which would indicate if crate is queryable). I think it's appropriate to specify what kind of check this is.

nslater: is "liveness" a standard term? I would have expected it to be "liveliness"

in any case, I am not sure it makes sense to disambiguate here unless we have multiple health checks to offer or that there's a reasonable chance of confusion

surely someone who's confused or doesn't know would follow the docs and learn what *we* mean by health checks?

also doesn't this allow us to expand the scope of our health checks in future?

nslater: just been linked to kube docs using "liveness" in which case okay

keep this :)

PR #78, opened 2018-02-08
Update resiliency page

mxm:
This is based on the latest state of https://github.com/elastic/elasticsearch/blob/master/docs/resiliency/index.asciidoc

### DIFF
--- a/docs/architecture/resilience.txt
+++ b/docs/architecture/resilience.txt
@@ -114,6 +114,14 @@ Here are some considerations:
    rare consistency or data-loss issues you might encounter when CrateDB
    experiences network or hardware failures (see next section).
 
+
+Nightly Jepsen tests
+--------------------
+
+The source code of these tests can be inspected in the official
+`Jepsen repository <https://github.com/jepsen-io/jepsen>`_.
+
+
 Known Issues
 ============
 
@@ -290,6 +298,54 @@ matches the lost update. This will break `Optimistic Concurrency Control
 <https://crate.io/docs/reference/sql/occ.html>`_.
 
 
+Replicas can fall out of sync when a primary shard fails
+--------------------------------------------------------
+
+.. raw:: html
+
+   <table class="summary">
+   <tr>
+    <td>Status</td>
+    <td>
+        Initial work to address this issue has been done by introducing
+        sequence numbers:
+        <a href="https://github.com/elastic/elasticsearch/issues/10708">#10708</a>
+    </td>
+   </tr>
+   <tr>
+    <td>Severity</td>
+    <td>Modest</td>
+   </tr>
+   <tr>
+    <td>Likelihood</td>
+    <td>Rare</td>
+   </tr>
+   <tr>
+    <td>Cause</td>
+    <td>Primary fails and in-flight writes are only written to a subset of its replicas</td>
+   </tr>
+   <tr>
+    <td>Workloads</td>
+    <td>Writes on replicated table</td>
+   </tr>
+   </table>
+
+.. rubric:: Scenario
+
+When a primary shard fails, a replica shard will be promoted to be the primary
+shard. If there is more than one replica shard, it is possible for the remaining
+replicas to be out of sync with the new primary shard. This is caused by
+operations that were in-flight when the primary shard failed and may not have
+been processed on all replica shards. Currently, the discrepancies are not
+repaired on primary promotion but instead would be repaired if replica shards
+are relocated (e.g., from hot to cold nodes); this does mean that the length of
+time which replicas can be out of sync with the primary shard is
+unbounded.
+
+.. rubric:: Consequence
+
+Stale data may be read from replicas.
+
 Fixed Issues
 ============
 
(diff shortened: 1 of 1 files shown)


docs/architecture/resilience.txt
```diff
@@ -114,6 +114,15 @@ Here are some considerations:
    rare consistency or data-loss issues you might encounter when CrateDB
    experiences network or hardware failures (see next section).
 
+
+Nightly Jepsen tests for dirty reads, lost updates, version divergence
+----------------------------------------------------------------------
+
+We run nightly tests of the Jepsen suite which test dirty reads, lost updates, and version divergence.
+The source code of these tests can be inspected in the official `Jepsen repository <https://github.com/jepsen-io/jepsen>`_.
+The tests are passing.
```
seut: that is not true, they are currently expected to fail which they do

docs/architecture/resilience.txt
```diff
@@ -290,6 +299,54 @@ matches the lost update. This will break `Optimistic Concurrency Control
...
+    <td>Severity</td>
+    <td>Modest</td>
+   </tr>
+   <tr>
+    <td>Likelihood</td>
+    <td>Rare</td>
+   </tr>
+   <tr>
+    <td>Cause</td>
+    <td>Primary fails and writes only to a subset of its replicas</td>
```
seut: would rephrase to: The primary shards fails while in-flight writes are not yet written to all replica shards.

docs/architecture/resilience.txt
mxm: I know we don't support all Jepsen tests but I thought the subset described here passes. Also the latest run passes: https://jenkins.crate.io/job/CrateDB/job/crate_jepsen_test/229/ The jepsen.log files in the artifact doesn't report any failure.

docs/architecture/resilience.txt
mxm: +1

docs/architecture/resilience.txt
seut: This job will pass even if the jepsen tests are failing (see https://jenkins.crate.io/job/CrateDB/job/crate_jepsen_test/200). Its not that easy to let them fail in a deterministic way. The main intention of this job is to ensure that our jepsen suite and tests are running without any issues against the current crate version despite of the test results.

mxm:
@seut Does this look good now?

docs/architecture/resilience.txt
```diff
@@ -114,6 +114,14 @@ Here are some considerations:
    rare consistency or data-loss issues you might encounter when CrateDB
    experiences network or hardware failures (see next section).
 
+
+Nightly Jepsen tests for dirty reads, lost updates, version divergence
```
seut: I would be so specific in what the jepsen suite is testing. this could change anytime and then we'd have update this document always..

docs/architecture/resilience.txt
```diff
@@ -437,6 +493,59 @@ the node with the original primary shard rejoins the cluster, CrateDB has no
 way of distinguishing correct and incorrect data, so that data replaced with
 data from the new primary shard.
 
+Make table creation resilient to closing and full cluster crashes
```
mfussenegger: Not sure if it makes sense to document issues that have been resolved quite a while ago?

docs/architecture/resilience.txt
mxm: I think we should be specific and it's ok to update this. Otherwise we can drop the entire paragraph.

docs/architecture/resilience.txt
mxm: I wanted to make this complete. Let's drop all resolved issues from the document then.

seut: I wouldn't drop existing old one as people may know them and want to see the update.

docs/architecture/resilience.txt
seut: Then lets remove it, I don't agree on being so specific here ;)

docs/architecture/resilience.txt
mxm: I specifically made sure to compare all issues with the ES document. I don't feel good with merging this in an incomplete state. I'm ok with removing the "fixed issues" section but not presenting it incomplete.

docs/architecture/resilience.txt
matriv: I like that we mention the Jepsen tests which somehow shows to our users that we care about the resiliency.
On the other hand I also don't like the specific mention of the tests. Personally I would keep it more abstract, but also I wouldn't like to remove the whole paragraph :-(

docs/architecture/resilience.txt
matriv: I don't have any strong opinion on that. If this issue was encountered in our older CrateDB versions, I don't mind to mention it here as fixed.

docs/architecture/resilience.txt
mxm: Removed the test details.

mxm:
@seut Made adjustments in line with your comments. Could you please have another look?

PR #91, opened 2018-04-24
add link to GitHub

nslater:
I am adding a block like this to all docs projects

### DIFF
--- a/docs/index.txt
+++ b/docs/index.txt
@@ -16,6 +16,11 @@ massive amounts of machine data in real-time.
    Check out the `CrateDB Reference`_ for low-level documentation, including
    release notes, configuration, syntax reference, and so on.
 
+.. SEEALSO::
+
+   The CrateDB Guide is an open source documentation project and is `hosted on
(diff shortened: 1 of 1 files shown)


docs/index.txt
```diff
@@ -16,6 +16,11 @@ massive amounts of machine data in real-time.
    Check out the `CrateDB Reference`_ for low-level documentation, including
    release notes, configuration, syntax reference, and so on.
 
+.. SEEALSO::
+
+   The CrateDB Guide is an open source documentation project and is `hosted on
```
Autophagy: Is 'hosted' the right word here? Since the hosting is provided by RTD, maybe 'available' would be more appropriate? What do you think?

nslater: I've thought about this and I can't think of a better one

"available" for me implies that it's downloadable from there or something, as a read-only thing. but I am subtly trying to encourage contribution

Autophagy: Fair enough!

PR #94, opened 2018-05-17
Restructure joins doc and add docs for HashJoin.

matriv:


docs/architecture/joins.txt
```diff
@@ -9,8 +9,8 @@ to work with large amounts of data.
 
 In this document we will present the following topics. First, an overview of
 the existing types of joins and algorithms provided. Then a description of how
-CrateDB implements them, and finally how CrateDB uses optimizations to work
-with huge datasets.
+CrateDB implements them along with the necessary optimizations, which allow to
```
andreidan: allow "them"/"us" to work (subject is missing)

docs/architecture/joins.txt
```diff
@@ -81,8 +83,12 @@ Joins in CrateDB
 ================
 
 CrateDB supports (a) CROSS JOIN, (b) INNER JOIN, (c) EQUI JOIN, (d) LEFT JOIN,
-(e) RIGHT JOIN and (f) FULL JOIN. To implement these, the nested loop join
-algorithm is implemented with a few optimizations.
+(e) RIGHT JOIN and (f) FULL JOIN. All of these join types are executed using
+the :ref:`nested loop join algorithm <joins_nested_loop>` along with some
+optimizations. :ref:`Equi Joins <joins_equi_join>` are executed using the
```
andreidan: maybe say "except the equi joins which are excuted using ..." ? 
(feels a bit weird to say " all are executed with NL" and then "equi joins are not")

docs/architecture/joins.txt
```diff
@@ -119,13 +127,13 @@ loop to acquire the data before being able to join. After finding the locations
...
-:ref:`joins_figure_2`).
+running a nested loop on the subset it has just received. Finally, these
+intermediate results are pushed to the original (planner) node to merge and
+return the results to the requesting client (see :ref:`joins_figure_2`).
 
-If the rows in the join result from or in another (nested) join or use a `table
-function`_ , the data is broadcast from (or to) different nodes directly.
+If the source rows of a join originate from another join result (joining more
+than 2 tables) or use a `table function`_ , the data is broadcast from (or to)
+different nodes directly.
```
andreidan: I think this is still a bit confusing. It's not really clear who the "different" nodes are

docs/architecture/joins.txt
```diff
@@ -119,13 +127,13 @@ loop to acquire the data before being able to join. After finding the locations
 of the required shards (which is done in the planning stage), the smaller data
 set (based on the row count) is broadcast amongst all the nodes holding the
 shards they are joined with. After that, each of the receiving nodes can start
-running a nested loop on the subset it has just received. Finally, the results
-are pushed to the original (planner) node to merge and return the results to
-the requesting client (see
-:ref:`joins_figure_2`).
+running a nested loop on the subset it has just received. Finally, these
+intermediate results are pushed to the original (planner) node to merge and
```
andreidan: shall we say "handler" node instead of the "planner" node? I think we generally use the "handler" term no?

docs/architecture/joins.txt
```diff
@@ -163,17 +159,123 @@ earlier:
...
+
+The Hash Join algorithm is used to execute certains types of joins in a more
+perfomant way than :ref:`Nested Loop <joins_nested_loop>`.
+
+Basic Algorithm
+...............
+
+The operation takes place in one node (the handler node to which the client is
+connected). The rows of the left relation of the join are read and a hashing
+algorithm is applied on the fields of that relation that participate in the
```
andreidan: maybe "on the fields of the relation which participate in the join" is a bit more clear (to avoid that that that)

docs/architecture/joins.txt
```diff
@@ -163,17 +159,123 @@ earlier:
...
+join condition. The hashing algorithm generates a hash value which is used to
+store every row of the left relation in the proper position in a `hash table`_.
+
+Then the rows of the right relation are read one-by-one and the same hashing
+algorithm is applied on the fields that participate in the join condition. The
+generated hash value is used to make a lookup in the `hash table`_. If no entry
+is found, the row is skipped and the processing continues with the next row from
+the right. If an entry is found, the join condition is validated (handling hash
```
andreidan: "the right relation".

docs/architecture/joins.txt
```diff
@@ -163,17 +159,123 @@ earlier:
...
+collisions) and on successful validation the combined tuple or left and right
```
andreidan: "combined tuple of left and right"

docs/architecture/joins.txt
matriv: I'm not sure if this is the correct meaning but how about:
```
the data is broadcast from the source to target nodes of directly.
```

docs/architecture/joins.txt
andreidan: Sounds good. Just a note "Special optimizations, according to the specific use cases, are applied to improve execution performance."

docs/architecture/joins.txt
andreidan: Yeah, I think this is better. Maybe we should specify who the source/target is?
Eg. 
`the data is broadcasted from the source join nodes directly to the nodes executing the current join.`

seut: I'm thinking about removing this completely as it may confuse more than it helps.

docs/architecture/joins.txt
```diff
@@ -163,17 +160,123 @@ earlier:
...
+rows of the left relation are processed and inserted into the `hash table`_
+until the block-size is reached. The operation then starts reading the rows of
+the right relation, process them one-by-one and performs the lookup and the
+join condition validation. Once all rows from the right relation are processed
+the `hash table`_ is re-initialized based on a new calculation of the block
+size and a new iteration starts until all rows of the left relation are
+processsed.
+
+With this algorithm the memory limitation is raised in expense of having to
+iterate over the rows of the right table multiple times.
```
seut: I think somewhere we should highlight that the `block hash join` will always be used rather than the normal `hash join`. Also maybe mention that we will try use a `distributed block hash join` when possible (no limit/offset).

docs/architecture/joins.txt
```diff
@@ -163,17 +160,123 @@ earlier:
...
+''''''''''''''''''''''''''
+
+Since the right table can be processed multiple times (number of rows from left
+/ block-size) the right table should be the smaller (in number of rows) of the
+two relations participating in the join. Therefore, if originally the right
+relation is larger than the left the query planner performs a switch to take
+advantage of this detail and execute the hash join with better performance.
+
+
+Distributed Hash Join
```
seut: indeed its a `Distributed Block Hash Join`, think we should mention that.

docs/architecture/joins.txt
```diff
@@ -163,17 +159,123 @@ earlier:
...
+Query Then Fetch
+................
+
+Join operations on large relation can be extremely slow especially if the join
+is executed with a :ref:`Nested Loop <joins_nested_loop>`. - which means that
+the runtime complexity grows quadratically (O(n*m)). Specifically for
+`Cross Joins`_ this results in large amounts of data sent over the network and
+loaded into memory at the handler node. CrateDB reduces the volume of data
+transferred by employing Query Then Fetch: First, filtering and ordering are
+applied (if possible where the data is located) to obtain the required document
```
mxm: Isn't this always local?

matriv:
@seut fixup pushed regarding your comments.

docs/architecture/joins.txt
```diff
@@ -163,17 +159,123 @@ earlier:
 Consequently, the number of rows is significantly reduced allowing the
 operation to complete much faster.
 
+.. _joins_hash_join:
+
+Hash Join
+---------
+
+The Hash Join algorithm is used to execute certains types of joins in a more
+perfomant way than :ref:`Nested Loop <joins_nested_loop>`.
```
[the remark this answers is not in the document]
matriv: it's written, than nested loop.

docs/architecture/joins.txt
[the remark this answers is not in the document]
matriv: already rephrased, please check again.

matriv:
@mxm also addressed your comments.

I will add the diagrams in a separate PR.

mxm:
Looks good! 🇬🇷

matriv:
great. thank you!

PR #100, opened 2018-06-25
add new join graphics

nslater:
opening this pr to collect changes


nslater:
this is just the initial version of these images @matriv, so no need to review just yet unless you want to. I have sent some requests for changes back to Lilly and I'll update the pr once I have the second version of these images + the missing one

matriv:
@nslater for https://github.com/crate/guide/blob/0afbde962f6d867da451b184b8e6f222992706e9/docs/architecture/hash-join.jpg
The data nodes are all 1 where there should be 1,2,...n
Also the decision arrows is gray and blue which doesn't show well the found/not found decision(green - red)

matriv:
https://github.com/crate/guide/blob/0afbde962f6d867da451b184b8e6f222992706e9/docs/architecture/nested-loop.jpg the middle block NestedLoop says L1 and L2 but it should say R1 and R2 (right table)

docs/architecture/joins.txt
```diff
@@ -26,13 +26,12 @@ elements appear in which join.
...
-.. figure:: joins.png
-   :scale: 70%
+.. figure:: joins.jpeg
    :align: center
 
-   Figure 1
+   Join Types
 
-   LEFT JOIN, RIGHT JOIN, INNER JOIN, OUTER JOIN, and CROSS JOIN of a set L and
+   From left to right, top to bottom: left join, right join, inner join, outer join, and CROSS JOIN of a set L and
```
matriv: IMHO I would keep the join types with capital letters, as CROSS JOIN is

matriv:
In this the shards of left and right are copy pasted in all 3 nodes but in the original diagram the shard distribution is different (more real-world like) https://github.com/crate/guide/blob/0afbde962f6d867da451b184b8e6f222992706e9/docs/architecture/distributed-hash-join.jpg

docs/architecture/joins.txt
nslater: hmm. in the documentation I am trying to keep a separation between `LEFT JOIN` as a specific piece of SQL syntax, an "left join" as a concept

does this make sense?

nslater:
@matriv thanks for the feedback, I have passed it on!

docs/architecture/joins.txt
matriv: I don't insist, it's just my preference, at least then make the cross join lowercase as well.

nslater:
@matriv okay updated the images

I am still waiting on the revised image per your feedback

nslater:
@matriv okay we have all the images in place. please re-review!

matriv:
@nslater for https://github.com/crate/guide/pull/100/files#diff-c060db4a7633a4776e9aa32db0361a57 The shard numbering is still wrong...
Rows of Left Sh are: 1,2,3,4,6,7 instead of 1,2,3,4,5,6
Rows of Right Sh are: 1,2,3,4,5,5,6,7,8 instead of 1,2,3,4,5,6,7,8,9

docs/architecture/joins.txt
matriv: This is not fixed (at least make CROSS JOIN lowercase)....

nslater:
@matriv okay thanks! I have notified Lilly

nslater:
@matriv everything is fixed now hopefully. can you check again?

docs/architecture/joins.txt
matriv: @nslater CROSS JOIN is still uppercase...

PR #101, opened 2018-06-25
add docs for config files

nslater:


docs/deployment/linux/debian.txt
```diff
@@ -78,6 +78,15 @@ Here, replace ``CHANNEL`` as above, and then, additionally, replace
...
 ``jessie``, or ``stretch``.
 
+Once that is done, update Apt:
+
+.. code-block:: sh
+
+   sh$ sudo apt-get update
+
+You should see a success message, indicating that the CrateDB release channel
+is working as expected.
```
chaudum: that the CrateDB packages have been added to the list of available packages.

docs/deployment/linux/debian.txt
```diff
@@ -169,6 +178,21 @@ In both instances, replace ``COMMAND`` with ``start``, ``stop``, ``restart``,
 Configuration
 =============
 
+Configuration Files
+-------------------
+
+The main CrateDB configuration files live in the ``/etc/crate`` directory.
+
+.. TIP::
```
chaudum: Not sure really, whether this should be a TIP. Changing an installation configuration provided by a package can be dangerous and lead to all sorts of problems. I would at least mention, that changing the configuration directory may cause unexpected behaviour.
If you want to tinker around, the simpler tarball installation should be used.

chaudum: If you want to see what files have been installed and where they are located, you could use ` dpkg-query -L crate`.

nslater:
@chaudum thanks for the feedback! can you re-check please?

also, this time I added info for red hat too. but I'm struggling to get a red hat environment set up locally to test with so my changes are my best guess

PR #104, opened 2018-08-07
Kubernetes docs

nslater:


autophagy:
Shall we also make refactoring the `scaling` section part of this? Since what is in the scaling section and in the deployment section are sort of duplicated?

nslater:
@autophagy yup! good idea!

PR #105, opened 2018-08-13
Refactor Kubernetes scaling documentation

autophagy:
I decided to stick with the already existing documentation, and strip out all the information that was not directly about scaling. Some of the information, like the section on persistent volumes, could be moved into the deployment guide. What do you think?

### DIFF
--- a/docs/scaling/kubernetes.txt
+++ b/docs/scaling/kubernetes.txt
@@ -1,294 +1,214 @@
-==========
-Kubernetes
-==========
+.. _scaling-kube:
 
-Kubernetes is an open-source system for automating deployment, scaling, and
-management of containerized applications that builds on 15 years of experience
-running production workflows at Google.
+=============================
+Scaling CrateDB on Kubernetes
+=============================
 
-Thanks to our :ref:`Docker image <cratedb_docker>`, managing your CrateDB
-cluster with Kubernetes requires only a few steps.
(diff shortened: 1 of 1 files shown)


nslater:
@autophagy I have pushed a fixup edit. can you take a look?

docs/scaling/kubernetes.txt
```diff
@@ -1,78 +1,210 @@
...
+    reflected in your `version control`_ system.
+
+.. _scaling-kube-vc:
 
-or by editing the value in the StatefulSet directly:
+Using Version Control
+---------------------
+
+If you want to version control your cluster configuration, you can edit the
+StatefulSet controller configuration file directly:
```
autophagy: Maybe worth mentioning here that the thing you're updating here is the `replicas` setting.

docs/scaling/kubernetes.txt
```diff
@@ -1,294 +1,210 @@
...
+    apiVersion: "apps/v1"
+    # We're using a StatefulSet controller.
+    kind: StatefulSet
+    metadata:
+      # This is the name used as a prefix for all pods in the set.
+      name: crate
+    spec:
+      serviceName: "crate-set"
+      # Our cluster has three nodes.
+      replicas: 3
```
autophagy: Also, perhaps this should be 4? Given the example above.

PR #107, opened 2018-09-17
refresh images

nslater:


nslater:
@emilywoods we have a designer atm so I'm taking the opportunity to update our docs to be more in line with our current site design!

PR #111, opened 2018-11-12
Mention that store.throttle only applies to CrateDB < 3.0

mfussenegger:


nslater:
I would reword it as:

> These settings were removed in CrateDB 3.0.

mfussenegger:
@nslater pushed a fixup

nslater:
ack! sorry for not getting back to you sooner. was on PTO for the end of last week

PR #114, opened 2018-11-28
fix Docker troubleshooting docs

nslater:


docs/deployment/containers/docker.txt
```diff
@@ -223,12 +223,21 @@ Here's how you would mount ``crate.yml``::
...
+have crashed or hanged inside the container without terminating.
+
+If you use `Docker Swarm`_ and you are experiencing trouble starting your Docker
+containers, try to deactivate the health check.
+
+You can do that like so:
+
+.. code-block:: yaml
+
+    disable: true
```
autophagy: Mm, this is unclear. I'd either include the full path down to the setting:
```
services:
 service:
    healthcheck:
     disable: true
```

Or just:
```
healthcheck:
  disable: true
```

PR #118, opened 2019-01-30
Mark USR2 signal node stop as deprecated

smakalias:
relates to crate/crate#8163

### DIFF
--- a/docs/admin/rolling-upgrade.txt
+++ b/docs/admin/rolling-upgrade.txt
@@ -126,6 +126,13 @@ Use the `SET`_ command to do so:
 Step 2: Graceful Stop
 ---------------------
 
+There are two ways of stopping a ``crate`` node:
+
+1. Use the ``ALTER CLUSTER DECOMMISSION`` statement.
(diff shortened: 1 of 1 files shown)


docs/admin/rolling-upgrade.txt
```diff
@@ -137,6 +137,11 @@ need to send the ``USR2`` signal::
   If your system uses SysVinit (e.g. Debian Wheezy), you can also use ``service
   crate graceful-stop``.
 
+.. WARNING::
+
+    Stopping a crate process using the ``USR2`` signal is deprecated
+    and will be removed in a future release.
```
mfussenegger: I think this should link back to the decomission statement. Or change the example altogether to use the decomission statement and mention the `USR2` in a note.

The deprecation should also be mentioned in the release notes.

smakalias:
@mfussenegger pushed a fixup mentioning 2 options for stopping a node with a deprecation warning for the 2nd.

mfussenegger:
@smakalias Can we merge this?

PR #121, opened 2019-02-12
Update linux installation instructions

mfussenegger:
- Removes instructions for distros based on Red Hat 6
 - Removes Debian 7 and 8. We only support the latest
 - Removes the intermediate Ubuntu releases, we only support the LTS
 versions
 - Removes the instructions about using `/etc/init.d/` on Debian 9. That
 was incorrect, is also using systemd.
 - Adapts the instruction on how to install JDK on Ubuntu < 16. There
 java 8 is not in the stock repository. (It is now recommending to
 install java 11)

### DIFF
--- a/docs/deployment/linux/debian.txt
+++ b/docs/deployment/linux/debian.txt
@@ -7,11 +7,7 @@ Run CrateDB on Debian GNU/Linux
 CrateDB maintains packages for the follow Debian versions:
 
 - `Stretch`_ (9.x)
-- `Jessie`_ (8.x)
-- `Wheezy`_ (7.x)
 
-.. _Wheezy: https://www.debian.org/releases/wheezy/
-.. _Jessie: https://www.debian.org/releases/jessie/
 .. _Stretch: https://www.debian.org/releases/stretch/
 
 This document will walk you through the process of installing and configuring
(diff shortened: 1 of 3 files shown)


docs/deployment/linux/ubuntu.txt
```diff
@@ -7,25 +7,23 @@ Run CrateDB on Ubuntu
...
 .. _Trusty Tahr: https://wiki.ubuntu.com/TrustyTahr/ReleaseNotes
 
 .. NOTE::
 
    CrateDB requires Java 8 or higher.
 
-   To run CrateDB on Ubuntu releases older than Trusty Tahr 14.10, you will need
-   to :ref:`install Java 8 from a 3rd party repository <debian-wheezy>`.
+   To run CrateDB on Ubuntu releases older than 16.04, you will need to install
+   Java from a 3rd party repository. This can be done by adding the openjdk ppa:
```
mechanomi: "third-party"

"openjdk PPA package"

docs/deployment/linux/red-hat.txt
```diff
@@ -78,14 +70,7 @@ With Red Hat Linux 7, you can control the ``crate`` service like so:
...
 
-With Red Hat Linux 6, you should use:
-
-.. code-block:: sh
-
-   sh$ sudo service crate COMMAND
-
-In both instances, replace ``COMMAND`` with ``start``, ``stop``, ``restart``,
-``status``, etc.
+Replace ``COMMAND`` with ``start``, ``stop``, ``restart``, ``status``, etc.
```
mechanomi: "Here, replace" ... "`status`, and so on."

docs/deployment/linux/debian.txt
```diff
@@ -166,14 +117,7 @@ so:
...
 
-With Debian Wheezy (9.x), you must do this instead:
-
-.. code-block:: sh
-
-    sh$ sudo /etc/init.d/crate COMMAND
-
-In both instances, replace ``COMMAND`` with ``start``, ``stop``, ``restart``,
-``status``, etc.
+Replace ``COMMAND`` with ``start``, ``stop``, ``restart``, ``status``, etc.
```
mechanomi: "Here, replace" ... "`status`, and so on."

(I know this wasn't in the original file, but we may as well improve it while we're here

docs/deployment/linux/ubuntu.txt
mfussenegger: Not sure if "package" is correct. It's a repository. PPA stand for Personal Package Archive, so `PPA package` sounds a bit like `HTTP Protocol`.

mechanomi: oh you're right yes! just "PPA"! thx

mfussenegger:
@mechanomi pushed a fixup

docs/deployment/linux/debian.txt
```diff
@@ -117,7 +117,8 @@ so:
 
     sh$ sudo systemctl COMMAND crate
 
-Replace ``COMMAND`` with ``start``, ``stop``, ``restart``, ``status``, etc.
+Here, replace ``COMMAND`` with ``start``, ``stop``, ``restart``, ``status`` and
```
mechanomi: eep. there should be a "," before the "and so on"

docs/deployment/linux/red-hat.txt
```diff
@@ -70,7 +70,8 @@ With Red Hat Linux 7, you can control the ``crate`` service like so:
 
    sh$ sudo systemctl COMMAND crate
 
-Replace ``COMMAND`` with ``start``, ``stop``, ``restart``, ``status``, etc.
+Here, replace ``COMMAND`` with ``start``, ``stop``, ``restart``, ``status`` and
```
mechanomi: same as above

PR #124, opened 2019-02-26
revamp tools page

mechanomi:


docs/tools/index.txt
```diff
@@ -5,69 +5,47 @@ Tools
...
--  `Website <http://www.metabase.com/>`__
+   streamsets
 
-.. _client tools: https://crate.io/a/category/client-tools/
+.. _blog post: https://crate.io/a/visualize-crate-data-with-metabase/
+.. _integrations: https://crate.io/a/category/client-tools/
+.. _Metabase: http://www.metabase.com/
+.. _our JDBC driver: https://crate.io/docs/reference/jdbc
+.. _Pentaho: http://community.pentaho.com
+.. _Petaho Kettle: https://github.com/pentaho/pentaho-kettle>
```
emilywoods: remove `>`?

docs/tools/index.txt
```diff
@@ -5,69 +5,47 @@ Tools
...
 
-Pentaho
--------
+- **StreamSets Data Collector**
 
-Pentaho prepares and blends data, delivering business analytics from any
-source. You can connect to CrateDB clusters by using the standalone version of
-`our JDBC driver <https://crate.io/docs/reference/jdbc/>`_.
+  The `StreamSets`_ Data Collector is a lightweight, powerful engine that
+  streams data in real time. Read `the manual`_ on how to use CrateDB as an
```
emilywoods: `Read the manual on how to use CrateDB as an origin` - does this suggest that there is a manual for using CrateDB as an origin and destination? 

Maybe `read the StreamSets manual where CrateDB can be set as an origin..`?

docs/tools/index.txt
mechanomi: good catch! thx!

docs/tools/index.txt
mechanomi: another good catch! I got the link wrong. pushed up a fixup that corrects the link text and the link URL

requirements.txt
```diff
@@ -15,7 +15,7 @@ commonmark==0.5.4
 docutils==0.13.1
 mock==1.0.1
 pillow==2.6.1
-Pygments==2.2.0
+Pygments==2.3.1
```
emilywoods: was this intended to change? i'd include a commit message + possibly separate commit for it if so

mechanomi: ugh no that wasn't meant to be included. taking it out. thanks!

PR #129, opened 2019-03-22
Add a select performance section for modulo based sampling

mfussenegger:


mfussenegger:
This depends on https://github.com/crate/crate/pull/8438 - maybe we should note here in the guide that `_docid` will only be available in 4.0?

docs/performance/selects.txt
```diff
@@ -0,0 +1,90 @@
...
+   +-----------+------------+--------------------+------------+
+
+You'll notice that the result changed slightly, but is still fairly close to
+the original result.
+
+Note that the ``% 10`` in this example was arbitrary. The higher the number,
+the fewer records will match the query and the less accurate the result will
+be, trading accuracy for more performance.
+
+
```
mfussenegger: Should we include some output like the following, so that users get a better intuition about the modulo?

```
cr> select _docid, value, _docid % 2 from measures limit 5;
+--------+-------+--------------+
| _docid | value | (_docid % 2) |
+--------+-------+--------------+
|      0 |  7745 |            0 |
|      1 |  4934 |            1 |
|      2 |  8315 |            0 |
|      3 |  9957 |            1 |
|      4 |  6380 |            0 |
+--------+-------+--------------+
```

docs/performance/selects.txt
```diff
@@ -0,0 +1,90 @@
...
+   |         4 |      10000 | 5013.052623224065  |          1 |
+   |         3 |      10000 | 4999.1253575025175 |          0 |
+   |         2 |      10000 | 5001.400379047543  |          0 |
+   |         1 |      10000 | 5035.220951927276  |          0 |
+   +-----------+------------+--------------------+------------+
+
+You'll notice that the result changed slightly, but is still fairly close to
+the original result.
+
+Note that the ``% 10`` in this example was arbitrary. The higher the number,
```
mfussenegger:
```suggestion
Note that the ``% 10`` in this example was arbitrary. It translates roughly to "Only consider every 10th row". The higher the number,
```

docs/performance/selects.txt
mechanomi: it can't hurt!

docs/performance/selects.txt
mechanomi: > It roughly translates to "only consider...

mechanomi:
@mfussenegger might be easier if I add an edit commit to your branch. is that ok?

mfussenegger:
> 
> 
> @mfussenegger might be easier if I add an edit commit to your branch. is that ok?

Sounds good

docs/performance/selects.txt
```diff
@@ -0,0 +1,90 @@
...
+
+A common use case for CrateDB is to do ``GROUP BY`` queries on tables for
+analytic purposes. Say you're selecting the ``avg``, ``max`` and ``min`` of
+some measurements over a billion records and group them by a device id.
+
+By default CrateDB processes all matching records, which may require a lot of
+processing power depending on the data set and the size of the CrateDB cluster.
+
+Similar products sometimes limit the amount of records that are processed for
+GROUP BY operations. That means the user might get inaccurate results, but for
```
lukas-bernert: I believe that "inaccurate results" sound a little too harsh?
How about "less accurate".
Or "simialar products trade accuracy for performance" ?

docs/performance/selects.txt
```diff
@@ -0,0 +1,88 @@
...
+
+It is common to do ``GROUP BY`` queries on tables for analytics purposes. For
+example, you might select the ``avg``, ``max``, and ``min`` of some
+measurements over a billion records and group them by device ID.
+
+By default, CrateDB processes all matching records. This may require a lot of
+processing power, depending on the data set and the size of the CrateDB
+cluster.
+
+Similar products sometimes limit the amount of records that are processed for
```
mechanomi: define/explain "similar products"

mechanomi: why do they limit the amount of processed records?

docs/performance/selects.txt
```diff
@@ -0,0 +1,88 @@
...
+``GROUP BY`` operations. That means the user might get inaccurate results, but
+for analytics this is often an acceptable trade-off.
+
+To emulate this down sampling behavior, users can filter on the ``_docid``
```
mechanomi: I think we're jumping in too early here and we might need to unpack this a bit so readers have more context/understanding

same is true for the rest of the doc so will hold off leaving further comments until the previous para is expanded and I have a better understanding myself

docs/performance/selects.txt
```diff
@@ -0,0 +1,88 @@
...
+the original result.
+
+Note that the ``% 10`` in this example was arbitrary. The higher the number,
+the fewer records will match the query and the less accurate the result will
+be. This trades accuracy for performance.
+
+.. NOTE::
+
+   The ``_docid`` system column exposes the internal document ID each document
+   has within a Lucene segment. The IDs are unique within a segment but not
```
mechanomi: "Lucene segment" needs linking

mfussenegger:
Can you elaborate a bit more on what you're expecting? There isn't much more to it. Processing all records takes a lot of time, processing fewer records takes less time. That's it.

mechanomi:
@mfussenegger see my previous/first comment in my first review. I don't understand this sentence: "Similar products sometimes limit the amount of records that are processed". what is a "product" and why does it limit the processed records

mfussenegger:
> "Similar products sometimes limit the amount of records that are processed". what is a "product" and why does it limit the processed records

products here is referring to the product "CrateDB" - similar products are our competitors.

I think the "why does it limit the processed records" is implicitly explained by the whole "trade accuracy for performance" part?

Not sure what the expectations towards the guide are, but the intention behind documenting this is for people who are already familiar somewhat with this approach of reducing accuracy and are specificly looking for doing that in CrateDB. It's not intended as a full fledged tutorial.

mechanomi:
@mfussenegger aha! "database products" is what I needed to contextualize and make sense of the rest of the text. I was thinking "products" was some sort of mathematical/internal term

I have provided a final set of edits that resolve all of my previous comments on this pr

mfussenegger:
It looks like my edits in https://github.com/crate/guide/commit/bff23178b0fcc3e78077fa344140bec91abae108#diff-28f5f1f949f42ed9211bb7e61348fdccR49 weren't included in the last edits. Do you think it is better to leave that section about using the `LIMIT` out of it?

PR #146, opened 2019-07-23
Update Docker guide to use Zen2 discovery settings

MarkusH:

CrateDB 4.0 switched to Zen2 as a discovery algorithm. As part of that, some configuration settings changed that needed to be updated

## Checklist

### DIFF
--- a/docs/deployment/containers/docker.txt
+++ b/docs/deployment/containers/docker.txt
@@ -100,7 +100,12 @@ So, let's create a second node, called ``crate02``, and configure it to ping
     sh$ docker run -d --name=crate02 \
           --net=crate -p 4202:4200 --env CRATE_HEAP_SIZE=2g \
           crate -Cnetwork.host=_site_ \
-                -Cdiscovery.zen.ping.unicast.hosts=crate01
+                -Cdiscovery.seed_hosts=crate01
+
+.. NOTE::
+
+    In versions of CrateDB prior to 4.0, you must use
+    ``discovery.zen.ping.unicast.hosts`` instead of ``discovery.seed_hosts``.
 
 Notice here how we also updated the port mapping, so that port ``4200`` on the
 container is mapped to ``4202`` on our host machine.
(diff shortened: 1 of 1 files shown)


docs/deployment/containers/docker.txt
```diff
@@ -100,7 +100,7 @@ So, let's create a second node, called ``crate02``, and configure it to ping
     sh$ docker run -d --name=crate02 \
           --net=crate -p 4202:4200 --env CRATE_HEAP_SIZE=2g \
           crate -Cnetwork.host=_site_ \
-                -Cdiscovery.zen.ping.unicast.hosts=crate01
```
mechanomi: agree with jordi

perhaps add something like this:

```
.. NOTE::

    In versions of CrateDB prior to 4.0, you must use ``discovery.zen.ping.unicast.hosts`` instead of ``discovery.seed_hosts``.

File: docs/deployment/containers/docker.txt [hunk 2]
```diff
@@ -147,18 +147,6 @@ the ``crate`` command::
 
    This will result in degraded performance.
 
-You can also start a single node without any bootstrap checks by passing the
```
mechanomi: why are we dropping this?

MarkusH: I couldn't find a way to disable bootstrap a node without checks. Neither could I find the setting in the docs. Thus I removed it.

MarkusH:
> It's also missing the initial master nodes setting. Will the nodes form a cluster properly with the settings as stated here now?

I got 3 containers to talk and synchronize as documented now. The only thing that I noticed were some warnings around `gateway.expected_nodes`. I suspect the `initial_master_nodes` isn't necessary or it worked without because I started the containers one after the other. When crate01 was running I started crate02 and it joined the cluster. I then started crate03 and it joined as well.

docs/deployment/containers/docker.txt
mfussenegger: The discovery.type setting should still cause CrateDB to not enforce the bootstrap checks.

I agree with Naomi, this shouldn't be dropped

mfussenegger:
> I got 3 containers to talk and synchronize as documented now.

Did you've some state from an earlier CrateDB around, or maybe this was somehow working because it was all running on the same node?

To form a proper cluster across multiple machines the `initial_master_nodes` setting is necessary.

mfussenegger:
I think this has been mostly covered with https://github.com/crate/guide/pull/152/files

Can we close this?

MarkusH:
I guess. Yes. Thanks.

PR #147, opened 2019-08-06
Update required max-number-of-threads value.

seut:
Since CrateDB 3.0.0, `4096` is required at least.


seut:
Already merged, seems like a GH hiccup.

PR #152, opened 2019-09-25
Improve the docker-compose example, as well as verify prev section

marregui:
## Checklist

### DIFF
--- a/docs/deployment/containers/docker.txt
+++ b/docs/deployment/containers/docker.txt
@@ -34,8 +34,13 @@ Quick start
 Creating a cluster
 ------------------
 
-To get started with CrateDB and Docker on your dev machine, start by creating a
-`user-defined network`_, like so::
+To get started with CrateDB and Docker, we will create a three node cluster
+on your dev machine. The cluster will run on a dedicated network and will
+require the first two nodes, ``crate01`` and ``crate02``, to vote which one
+is the master. The third node, ``crate03``, will simply join the cluster
+with no vote.
+
+To create the `user-defined network`_, run the command::
 
     sh$ docker network create crate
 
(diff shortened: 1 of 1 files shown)


marregui:
In relation with 
https://stackoverflow.com/questions/58051469/setting-up-cratedb-4-0-4-cluster

docs/deployment/containers/docker.txt
```diff
@@ -192,26 +208,55 @@ Read about Docker Compose specifics `here <https://docs.docker.com/compose/>`_.
...
+      image: crate
+      container_name: crate01
+      hostname: crate01
+      ports:
+        - 4201:4200
+      net: crate
+      volumes:
+        - /tmp/crate/01:/data
+      command: >
+        crate -Cnetwork.host=_site_
```
seut: also this node should have its seed_hosts and initial_master_nodes configured

marregui: Pointing to itself?, or the other two nodes?. Should each node be aware of the other two?

seut: yes each node should be aware of all other nodes as the order in how they are started may not be deterministic and the configuration should not rely on that

docs/deployment/containers/docker.txt
```diff
@@ -219,6 +219,8 @@ Read about Docker Compose specifics `here <https://docs.docker.com/compose/>`_.
         - /tmp/crate/01:/data
       command: >
         crate -Cnetwork.host=_site_
+        -Cdiscovery.seed_hosts=crate02,crate03
+        -Ccluster.initial_master_nodes=crate02,crate03
```
seut: also the local node should be included in `initial_master_nodes` to also be part of the initial master voting.

marregui: I suppose for all nodes?

seut: yep

docs/deployment/containers/docker.txt
```diff
@@ -95,12 +105,15 @@ user-defined ``crate`` network, so any subsequent nodes can access this node
...
+so::
 
     sh$ docker run -d --name=crate02 \
-          --net=crate -p 4202:4200 --env CRATE_HEAP_SIZE=2g \
+          --net=crate -p 4202:4200 \
+          --env CRATE_HEAP_SIZE=2g \
           crate -Cnetwork.host=_site_ \
-                -Cdiscovery.seed_hosts=crate01
+                -Cdiscovery.seed_hosts=crate01 \
+                -Ccluster.initial_master_nodes=crate01
```
seut: `crate02` should be also added to the `initial_master_nodes`

seut:
@marregui The link in my recent comment was wrong, updated it, please address.

docs/deployment/containers/docker.txt
```diff
@@ -101,40 +113,50 @@ You can visit the admin UI in your browser with this URL:
...
+                -Ccluster.initial_master_nodes=crate01,crate02
+
+Notice here that:
 
-Notice here how we also updated the port mapping, so that port ``4200`` on the
-container is mapped to ``4202`` on our host machine.
+- We also updated the port mapping, so that port ``4200`` on the container
+  is mapped to ``4202`` on our host machine
+- We have set parameter ``discovery.seed_hosts``, which defines the list of
+  master-eligible nodes. By default, if the parameter is not set, local ports
```
seut: "... which defines the list of discovery nodes"

docs/deployment/containers/docker.txt
```diff
@@ -101,40 +113,50 @@ You can visit the admin UI in your browser with this URL:
...
+  in the range 4300 to 4400 will be scanned. Otherwise, the format is a comma
+  separated list of ``host:port`` entries, where port defaults to setting
+  ``transport.tcp.port``.
```
seut: Maybe add explanation of `cluster.initial_master_nodes` here as well?

marregui:
I believe the flow now is correct. At first, a node is created with a self reference for initial master node. So crate01 becomes master on a single node cluster. Then we decide to extend the cluster by adding a second node, using crate01 as seed and both crate01 and crate02 as master voters. 
The way the third node is defined, it cannot vote in a master election. I hope this is correct.

What was a little misguiding for me is the fact that, unless explicitly defined, nodes take a random name. Maybe, as usually we only run a single node per host, nodes should take, by default, the host's name, unless while scanning 4300... another node responds, in which case perhaps subsequent nodes could be called a random name. This way, in docker examples, node and container name would match, without having to define two separate things?

docs/deployment/containers/docker.txt
```diff
@@ -136,10 +136,13 @@ Notice here that:
 - We also updated the port mapping, so that port ``4200`` on the container
   is mapped to ``4202`` on our host machine
 - We have set parameter ``discovery.seed_hosts``, which defines the list of
-  master-eligible nodes. By default, if the parameter is not set, local ports
+  discovery nodes. By default, if the parameter is not set, local ports
   in the range 4300 to 4400 will be scanned. Otherwise, the format is a comma
   separated list of ``host:port`` entries, where port defaults to setting
-  ``transport.tcp.port``.
+  ``transport.tcp.port``
+- ``cluster.initial_master_nodes``: Is not set to both nodes crate01 and
```
seut: Is this a typo and should be "Is now set.." instead?

PR #158, opened 2020-03-04
Update memory configuration section

mfussenegger:
The recommendation of 50% heap is outdated and doesn't take into
consideration that a lot of structures are now memory-mapped and
therefore no longer loaded into heap.

This extends the section by elaborating a bit on the trade-off between
small and big heap and recommends 25% of heap as a starting value.

### DIFF
--- a/docs/performance/memory.txt
+++ b/docs/performance/memory.txt
@@ -4,15 +4,39 @@
 Memory configuration
 ====================
 
-In general, you should allocate as much memory as possible to CrateDB.
+CrateDB is a Java application running on top of a Java Virtual Machine (JVM).
+For optimal performance you must configure the amount of memory that is
+available to the JVM for **heap** allocations. The **heap** is a memory region
+used for allocations of objects. For example, if you invoke a ``SELECT``
+statement, parts of the result set are temporarily allocated in the **heap**.
 
-The amount of memory made allocated is called the *heap*.
+The amount of memory that CrateDB can use for heap allocations is set using the
+`CRATE_HEAP_SIZE`_ environment variable.
 
-If the heap is too small, some memory-intensive queries (such as those which
-need to hold many rows in memory) may fail with ``OutOfMemory`` exceptions.
+The right memory configuration depends on your workloads.
 
-However, there are two important limits on how large you should configure the
-heap.
+Consider the following when determining the right value:
+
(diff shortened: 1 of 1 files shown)


mfussenegger:
fyi @joemoe

docs/performance/memory.txt
```diff
@@ -4,15 +4,40 @@
...
-need to hold many rows in memory) may fail with ``OutOfMemory`` exceptions.
+The right memory configuration depends on your workloads. Consider the
+following when determining the right value:
+
+The JVM has automatic memory management and frees up memory using `Garbage
+Collection`_ (GC). CrateDB, as of version 4.1, defaults to use a garbage
+collection implementation called `G1GC`_.
+
+This implementation performs well for heap sizes ranging from several GB to ten
+or more. It tries to provide the best balance between latency and throughput.
```
mechanomi: pull these paras into one, I think. the content is strongly co-related

docs/performance/memory.txt
```diff
@@ -4,15 +4,40 @@
...
+
+Still, the garbage collection times **increase** with a bigger heap, leading to
+**increased** latency.
+
+Therefore, your heap size shouldn't be too large.
```
mechanomi: same here. would pull these into a single para, for the same reason

docs/performance/memory.txt
```diff
@@ -4,15 +4,40 @@
...
+statement, parts of the result set are temporarily allocated in the **heap**.
 
-The amount of memory made allocated is called the *heap*.
+The amount of memory that CrateDB can use for heap allocations is set using the
+`CRATE_HEAP_SIZE`_ environment variable.
 
-If the heap is too small, some memory-intensive queries (such as those which
-need to hold many rows in memory) may fail with ``OutOfMemory`` exceptions.
+The right memory configuration depends on your workloads. Consider the
+following when determining the right value:
```
mechanomi: I would maybe pull the "Consider..." sentence into a new para, and then, because you're using a colon to introduce a list, I would format the next few paras as an unordered list

I would end the list with the last item being "CrateDB also uses Memory mapped files..."

mfussenegger: Pushed a fixup with the changes, but I'm not sure if I understood you right :)

PR #159, opened 2020-03-05
Add tutorial on how to remove nodes from a cluster

marregui:
This tutorial is a roundabout exemplification of the use of the [crate-node](https://crate.io/docs/crate/reference/en/latest/cli-tools.html?highlight=crate%20tool#crate-node) tool to detach and bootstrap nodes.

First a cluster is setup, by pointing the reader to existing tutorials, and by elaborating on how to build CrateDB from sources and test locally rather than on a production cluster.  Then some data is loaded, with references to the ISS tutorial, which is awesome. But also, suggesting alternative ways. Finally the 3 node cluster is made to be a single node cluster.


docs/scaling/downscaling.txt
```diff
@@ -0,0 +1,231 @@
...
+In this guide we:
+
+- Create a Vanilla cluster.
+- Add some data to it.
+- Downscale it to a single node cluster.
+
+
+.. _scaling-down-starting-vanilla-cluster:
+
+Starting a Vanilla cluster
```
matthijskrul: Is "Vanilla" a technical term here, or just meant in the sense of "generic"? If the latter I'd write it lowercase "vanilla"

marregui:
Hola @mechanomi ,

I was going to work on this guide, however there is this ``blocked`` label. Should I get the guide to merge point, or are the contents not relevant anymore?

Thank you

mechanomi:
@marregui the blocked label was to indicate to myself that you are blocked on this. but I changed how we're organizing the project board so I removed it

will get back to you on this soon

marregui:
Excellent thank you. I have pushed the latest version, so now it is down to PR reviewing amending.

mechanomi:
@marregui heya. I've taken a look! I have some high-level comments

---

I am not comfortable with the idea of shipping scripts in a directory that the reader has to clone. for a how-to guide, the scripts should be embedded in the text

however, it is fine to link to pastebin versions of them for ease of use. our current way of doing this is to use a pastebin service that is most appropriate to the type of code being used. the critical thing here, though, is that the full code is there in the doc. the pastebin is doesn't have anything extra in it. this allows us to re-create pastebin links if we want to without risking losing any information

also: please don't use GitHub to this, or any other service that ties the paste to a specific username. we want to avoid the situation where an individual author disables their account on a third-party service some time in the future and takes down a bunch of pastes that we are using

so my first request is this: can you rewrite this so that you introduce each script as you go along. explain what it does, how it can be used, etc. fit it into the "story" you're telling in the how-to

for some of the scripts, I think you should consider dropping them entirely. for example, the one that installs CrateDB. a better solution here is to link to our existing installation guides or deployment guides and let the user figure this out for themselves. we should keep the how-to guide as tight and focused as possible

I also don't think it's necessary to include the complete configuration for each node. we should assume a default configuration and only give specific example configuration snippets when the user is required to make alterations to the default config

---

can you please restructure the how-to so that you don't introduce all the scripts at the beginning. the scripts, or config snippets, or code snippets should appear in the text at the point where they are needed. this helps keep the how-to readable and formats it more like a story (which is easier to read)

---

for the sample data, I would suggest two alternative methods:

1. either point the reader at our tutorials showing them how to generate mock time series data using ISS telemetry information

2. show them how to use the cr8 tool (see https://crate.io/docs/crate/howtos/en/latest/performance/inserts/testing.html)

---

it's possible that this how-to becomes quite long with the changes I have requested. that's fine. as we continue to review and edit we can make the decision to split it up into multiple parts if that makes sense

marregui:
thank you!, sure, will do as you request.

curiosity, is pastebin this: https://pastebin.com/ ?

mechanomi:
@marregui that's one pastebin. but you should look around. there are other options. we should pick the one that is most used by the community (whichever community the doc is targeting)

marregui:
Hi @mechanomi 

I have deleted the scripts and inlined them where relevant. I have also revisited the prose to include hooks to other guides/documentation.

mechanomi:
@marregui can you let me know explicitly when you'd like me to take another look, please? I'm not sure from the comment you left. thanks!

marregui:
hi @mechanomi could you please have another look? thank you in advance.

marregui:
hi, I will away next week. I was wondering about the chances of merging the PR this week? thank you in advance

mechanomi:
@marregui heya. sorry, I've been sick this week. I hope to have a second review ready for you by the time you're back from vacation

marregui:
Hi @mechanomi I hope you are feeling better. I am back.

mechanomi:
@marregui there appears to be a downscaling.rst and a downscaling.txt file added. I assume that one of them is the file you want to add and the other is an older version of the same file. can you remove the older version? or indicate to me which one is the newest version so I know which file to edit

thank you!

marregui:
@mechanomi apologies for this, I have amended it. The correct document is *.rst.

mechanomi:
okay great!! thank you. this is looking much better now

since you started on this tutorial, the multi node cluster tutorial was updated quite significantly. check it out: https://crate.io/docs/crate/howtos/en/latest/clustering/multi-node-setup.html
can you update your tutorial so that it directs people there and uses the setup that is documented in that tutorial

if you need to suggest specific config changes, that's fine. but they should be config changes that can be applied to the cluster that the reader has set up following the previous tutorial

you can do away with the whole "Installing from sources" section. let's just provide the reader with one and only one way to do everything. and in this instance, we should direct them to the multi node cluster tutorial and work with the results of that. doing it this way will keep the tutorial easy to follow, easy to understand, and short

then, for thee "Adding some data" section, let's go a similar route. you have already linked to the "generate time series tutorial", which is great. but then the next section goes on to detail how to generate and work with a different type of mock data

for the same reasons as before, I think it's probably best if we take out the stuff to do with csv logs. that includes the script to handle the data as well as the tables/queries

that should simplify that whole section as it's mostly just deleting stuff

for the "Exploring the Data" section, could you rework this so that you're exploring the data generated by the "generate time series data" tutorial running on the multi node cluster set up by the previous tutorial? that way, the tutorial you are writing is building on both of those previous tutorials

this will include switching away from working with the logs table to working with the iss table

hopefully, all of the above is relatively straight forward as it mostly involves deleting and/or slightly adapting what you already have. the real meat of the tutorial is the last section where you demonstrate how to downscale the cluster. and hopefully, the changes I am suggesting don't substantially affect that bit

what do you think?

marregui:
Thank you for the feedback!. All sounds very reasonable, and I will proceed as you describe.

docs/clustering/downscaling.rst
```diff
@@ -0,0 +1,360 @@
...
+To start each node ``[n1, n2, n3]`` you can run a script (once per node) similar to:
+
+::
+
+    node_name=$1
+
+    path_conf="$(pwd)/conf/$node_name"
+    path_home=$(pwd)/crate
+    path_data=$(pwd)/data/$node_name
+    path_snapshots=$(pwd)/data/$node_name/snapshots
```
seut, line 120: In this tutorial no snapshots are used at all, so this is not needed.

docs/clustering/downscaling.rst
```diff
@@ -0,0 +1,360 @@
...
+
+    export CRATE_HOME=$path_home
+    if [ -z "$CRATE_HEAP_SIZE" ]; then
+        export CRATE_HEAP_SIZE="8G"
+    fi
+
+    ./crate/bin/crate -Cpath.conf=$path_conf \
+                      -Cpath.data=$path_data \
+                      -Cpath.repo=$path_snapshots \
```
seut, line 129: as said, not needed as no snapshots are used

docs/clustering/downscaling.rst
```diff
@@ -0,0 +1,360 @@
...
+    path_data=$(pwd)/data/$node_name
+    path_snapshots=$(pwd)/data/$node_name/snapshots
+
+    export CRATE_HOME=$path_home
+    if [ -z "$CRATE_HEAP_SIZE" ]; then
+        export CRATE_HEAP_SIZE="8G"
+    fi
+
+    ./crate/bin/crate -Cpath.conf=$path_conf \
+                      -Cpath.data=$path_data \
```
seut, line 128: As wrote above, this setting should be part of the config and thus can be remove here.

docs/clustering/downscaling.rst
```diff
@@ -0,0 +1,360 @@
...
+                      -Cpath.repo=$path_snapshots \
+                      -Cnode.name=<NODE_NAME> \
+                      -Ctransport.tcp.port=<PORT>
```
seut, line 131: both settings are defined inside the `crate.yml` template above and thus not needed.

docs/clustering/downscaling.rst
```diff
@@ -0,0 +1,360 @@
...
+here is how each operation is invoked from the command line:
+
+::
+
+    ./crate/bin/crate-node detach-cluster -Cpath.home=$(pwd)/crate \
+                                          -Cpath.conf=$path_conf \
+                                          -Cpath.data=$path_data \
+                                          -Cpath.repo=$path_snapshots \
+                                          -Cnode.name=<NODE_NAME> \
+                                          -Ctransport.tcp.port=<PORT>
```
seut, line 309: These settings are not needed by the crate-node tool (and also defined inside the template, or not needed at all like the repo.path)

docs/clustering/downscaling.rst
```diff
@@ -0,0 +1,360 @@
...
+When replication is completed, you need to take down all the nodes in the cluster,
+as you are going to externally affect is state by means of the crate-node-tool_. You
+will first detach nodes [n2, n3], and then will bootstrap node [n1]. For convenience
+here is how each operation is invoked from the command line:
+
+::
+
+    ./crate/bin/crate-node detach-cluster -Cpath.home=$(pwd)/crate \
+                                          -Cpath.conf=$path_conf \
+                                          -Cpath.data=$path_data \
```
seut, line 306: This setting must point to the concrete data path of each node, this should be explained/highlighted.

docs/clustering/downscaling.rst
```diff
@@ -0,0 +1,360 @@
...
+                                          -Cnode.name=<NODE_NAME> \
+                                          -Ctransport.tcp.port=<PORT>
+
+
+    ./crate/bin/crate-node unsafe-bootstrap -Cpath.home=$(pwd)/crate \
+                                          -Cpath.conf=$path_conf \
+                                          -Cpath.data=$path_data \
+                                          -Cpath.repo=$path_snapshots \
+                                          -Cnode.name=<NODE_NAME> \
+                                          -Ctransport.tcp.port=<PORT>
```
seut, line 317: Why not add here the concrete steps/ shell commands to execute to scale down to 1 node?
Like:
 - detach node 3 from the cluster:
   `... detach-cluster ... `
 - detach node 2 from the cluster:
   `... detach-cluster ... `
 - tell node 1 that it  should forget  about node 2 and 3:
   `... unsafe-bootstrap ...`

I think this is one of the most interesting section for this tutorial

docs/clustering/downscaling.rst
```diff
@@ -0,0 +1,360 @@
...
+
+
+The best practice is to select the node that was master in the cluster, to be the single
+node in the new cluster, as then we know it had the latest version of the cluster state.
```
seut, line 321: How do I know which node was the master? An example/description how to gather that information would be very helpful.

docs/clustering/downscaling.rst
```diff
@@ -0,0 +1,360 @@
...
+To set it up I recommend any of our `deployment guides`_, using this template for each
+of the nodes:
+
+::
+
+   cluster.name: vanilla
+   node.name: <NODE_NAME>
+   network.host: _local_
+   node.max_local_storage_nodes: 3
+   stats.service.interval: 0
```
seut, line 35: why is this setting relevant here?

docs/clustering/downscaling.rst
```diff
@@ -0,0 +1,360 @@
...
+For this tutorial, we are running in a single host so the cluster state is more or less
+guaranteed to be consistent across all nodes. In principle, however, the cluster could
+be running across multiple hosts, and then we would want the master node to become the
+new single node cluster.
+
+The new configuration for the single node:
+
+   ::
+
+     cluster.name: simple   # don't really need to change this
```
seut, line 331: why is the cluster name now different and why don't I need to change it?
```suggestion
     cluster.name: vanilla
```

docs/clustering/downscaling.rst
```diff
@@ -0,0 +1,360 @@
...
+     node.name: n1
+     stats.service.interval: 0
```
seut, line 333: again, this setting is not relevant, so I'd suggest to drop it.

docs/clustering/downscaling.rst
```diff
@@ -0,0 +1,360 @@
...
+   cluster.name: vanilla
+   node.name: <NODE_NAME>
+   network.host: _local_
+   node.max_local_storage_nodes: 3
+   stats.service.interval: 0
+
+   http.cors.enabled: true
+   http.cors.allow-origin: "*"
+
+   transport.tcp.port: <PORT>
```
seut, line 40: why is it needed to define this setting? afaik the default works fine.
Otherwise the settings for `discovery.seed_hosts` and `cluster.initial_master_nodes` would have to be adjusted as well.

docs/clustering/downscaling.rst
```diff
@@ -0,0 +1,360 @@
...
+   discovery.seed_hosts:
+       - 127.0.0.1:4301
+       - 127.0.0.1:4302
+       - 127.0.0.1:4303
+   cluster.initial_master_nodes:
+       - 127.0.0.1:4301
+       - 127.0.0.1:4302
+       - 127.0.0.1:4303
+
+where ``<PORT>`` is one of ``[4301, 4302, 4303]``, and ``<NODE_NAME>`` one of ``[n1, n1, n3]``.
```
seut, line 52: As said, no need for `<PORT>`.

docs/clustering/downscaling.rst
```diff
@@ -0,0 +1,360 @@
...
+     cluster.name: simple   # don't really need to change this
+     node.name: n1
+     stats.service.interval: 0
+     network.host: _local_
+     node.max_local_storage_nodes: 1
+
+     http.cors.enabled: true
+     http.cors.allow-origin: "*"
+
+     transport.tcp.port: 4301
```
seut, line 340: as said, not needed.

seut (commented):
Left a couple of comments. I'd suggest that someone which is not involved works through this to validate that all is understandable and works.

marregui:
Hi,

Thank you for the comments and advise. This tutorial was written a long time ago as an answer to a user's question [in community](https://community.crate.io/t/from-three-node-cluster-to-single-node/326). He was satisfied, he learned, and since then, a whole lot of content was produced and added without having any consideration towards the content in this tutorial. I think having to morph it entirely into something I never planned, specially after the fact (user is happy) seems wasteful to me. Thus I would like to discontinue work on this content and @mechanomi can decide whether to close, or amend/merge.

Thank you again.

PR #161, opened 2020-05-11
Fix build issues

autophagy:
This commit fixes build issues that were introduced with 1ebe6ad6fca6405fcf1c993ba3f414748e330265

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -1,3 +1,3 @@
 from crate.theme.rtd.conf.crate_howtos import *
 
-exclude_patterns = ["out/**"]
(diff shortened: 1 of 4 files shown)


docs/conf.py
```diff
@@ -1,3 +1,3 @@
 from crate.theme.rtd.conf.crate_howtos import *
 
-exclude_patterns = ["out/**"]
+exclude_patterns = ["out/**", "README.rst"]
```
mechanomi, line 3: minor, optional: this can be left in. once this build system is upgraded to use the docs utils this entire line will be removed. for the time being, it's harmless

PR #199, opened 2020-06-25
Add reference arhictecture document for PowerBI Desktop

autophagy:
Also fixes the failing build that's been happening for a week.

This currently uses the ODBC driver. We had hoped to write this with just using
the standard PostgreSQL driver bundled with Power BI, but at the moment that
is unavailable. We can update this document when that is the case.


docs/reference-architectures/powerbi.rst
```diff
@@ -0,0 +1,182 @@
...
+   :align: center
+
+You can hit *Test* to verify that the DSN is configured correctly. Once this
+is done, hit *Save* to save the DSN.
+
+
+Connecting Power BI to CrateDB
+------------------------------
+
+Once you have created the DSN, open up Power BI and click 'Get Data' from the
```
matthijskrul, line 137: You should be consistent about how you represent things to click and interact with: either all 'like so' or all _like so_

autophagy: Which do you think works best? I think i'd prefer *like this* but I don't know if we have a standard.

matthijskrul: Yeah in the Cloud docs I've generally used italics for that

PR #212, opened 2020-07-07
Fix slug, labels, TOC, trailing whitespace

mechanomi:
address comment on https://github.com/crate/crate-howtos/pull/207, plus bonus fixes

### DIFF
--- a/docs/troubleshooting/docker-jcmd.rst
+++ b/docs/troubleshooting/docker-jcmd.rst
@@ -1,8 +1,8 @@
 .. _jcmd-docker:
 
-====================================
-Troubleshooting Docker with ``jcmd``
-====================================
+===============================================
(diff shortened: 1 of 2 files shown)


docs/troubleshooting/docker-jcmd.rst
```diff
@@ -1,27 +1,34 @@
-============================================================
-Troubleshooting the ``crate`` Docker container with ``jcmd``
-============================================================
+.. _jcmd-docker:
+
+====================================
+Troubleshooting Docker with ``jcmd``
```
mfussenegger, line 4: This would be now a bit misleading as it's not troubleshooting docker, but it's about troubleshooting CrateDB running in a docker container.

mechanomi: @mfussenegger pushed a fixup

PR #225, opened 2020-07-17
RPM: Use correct repo definition for CrateDB >= 4.2.0

kovrus:

Starting with 4.2.0 we have to build platform-specific
rmp packages. That is the result of binding the platform-specific
JDK with CrateDB.

## Checklist

### DIFF
--- a/docs/deployment/linux/red-hat.rst
+++ b/docs/deployment/linux/red-hat.rst
@@ -34,6 +34,12 @@ For Red Hat Linux 7, run:
 
 .. code-block:: sh
 
+   sh$ sudo rpm -Uvh https://cdn.crate.io/downloads/yum/7/x86_64/crate-release-7.0-1.x86_64.rpm
+
+For CrateDB versions < 4.2.0, run:
+
+.. code-block:: sh
+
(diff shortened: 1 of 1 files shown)


docs/deployment/linux/red-hat.rst
```diff
@@ -34,7 +34,7 @@ For Red Hat Linux 7, run:
 
 .. code-block:: sh
 
-   sh$ sudo rpm -Uvh https://cdn.crate.io/downloads/yum/7/noarch/crate-release-7.0-1.noarch.rpm
+   sh$ sudo rpm -Uvh https://cdn.crate.io/downloads/yum/7/x86_64/crate-release-7.0-1.x86_64.rpm
```
mfussenegger, line 37: Should we also keep the `noarch` version in a note below, with the hint that that's the variant for CrateDB 4.1?

kovrus: I thought that in 4.1 doc it will be still noarch, but maybe let's add the note 👍

mfussenegger: Afaik the docs of the howto repo are not versioned like the crate/crate docs are.

mechanomi: only docs inside software project repos get versioned

PR #232, opened 2020-08-10
Update the parallel insert example + documentation

seut:
The example was wrong in that case that each concurrent insert must use a dedicated JDBC Connection instance as a Connection is not thread-safe.
Also, the usage of a pooled DataSource like HikariCP is highly recommended.

### DIFF
--- a/docs/performance/inserts/parallel.rst
+++ b/docs/performance/inserts/parallel.rst
@@ -24,6 +24,7 @@ response before sending another insert.
    :ref:`inserts_bulk_operations` are a good fit. In many cases, you will see
    even better performance from bulk opperations.
 
+
 .. rubric:: Table of contents
 
 .. contents::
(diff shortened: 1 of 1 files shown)


docs/performance/inserts/parallel.rst
```diff
@@ -33,19 +34,31 @@ Example
...
 
 .. code-block:: java
 
+    HikariDataSource ds = new HikariDataSource();
+    ds.setJdbcUrl("crate://localhost:5432/doc?user=crate");
+
     IntStream.iterate(0, i -> i + 2)
         .limit(1000)
         .forEach(i -> {
             CompletableFuture<Integer> insertFuture =
```
mfussenegger, line 57: Do you think it would also make sense to include a `CompletableFuture.allOf(..)` to show how to wait for all operations to be finished?

docs/performance/inserts/parallel.rst
```diff
@@ -33,35 +34,55 @@ Example
...
 Suppose we have a stream of data we want to persist into CrateDB.
-You can parallelize this in Java using a
-`CompletableFuture`_ object, like so:
+You can parallelize this in Java using a `CompletableFuture`_ object in order to
+issue insert statements concurrently (non-blocking).
+
+As the java JDBC `Connection`_ is not thread-safe, each insert must use a
+dedicated connection.
+
+A common solution to avoid over-instantiating of connection objects is the usage
```
mechanomi, line 43:
```suggestion
A common solution to avoid over-instantiating connection objects is the usage
```

docs/performance/inserts/parallel.rst
```diff
@@ -33,35 +34,55 @@ Example
...
+of a pooled JDBC `DataSource`_ like `HikariCP`_, `vibur-dbcp`_, `commons-dbcp`_
+etc.
```
mechanomi, line 45:
```suggestion
A common solution to avoid over-instantiating connection objects is the use of
a pooled JDBC `DataSource`_ like `HikariCP`_, `vibur-dbcp`_, or
`commons-dbcp`_.
```

docs/performance/inserts/parallel.rst
```diff
@@ -33,35 +34,55 @@ Example
...
+
+In the following example, `HikariCP`_ is used as a `DataSource`_.
```
mechanomi, line 47:
```suggestion
In the following example, `HikariCP`_ is used as a `DataSource`_:
```

PR #236, opened 2020-08-18
Update list of Debian distributions

mechanomi:


infoverload (approved):

docs/deployment/linux/debian.rst
```diff
@@ -4,11 +4,12 @@
...
 
-CrateDB maintains packages for the follow Debian versions:
+CrateDB provides packages for the following Debian versions:
 
+- `Buster`_ (10.x)
 - `Stretch`_ (9.x)
-
-.. _Stretch: https://www.debian.org/releases/stretch/
+- `Jessie`_ (8.x)
+- `Wheezy`_ (7.x)
```
mfussenegger, line 12:
```suggestion
```

We dropped Jessie and Wheezy and are currently only building packages for 9.x and 10.x

mechanomi: @mfussenegger as a part of this change, I checked our APT repositories and we are still hosting the packages as far as I can tell (see https://cdn.crate.io/downloads/apt/stable/dists/). which is why I changed the wording to "provide" instead of "maintain" (so as to not falsely indicate that the existence of these packages means they are going to be updated)

have I misunderstood? should we clean up our apt repository?

mfussenegger: > should we clean up our apt repository?

No, we don't want to make it impossible for people to re-install a working setup.

Maybe we just make it more explicit and distinguish between actively maintained (receiving new packages), that for older distributions we provide packages up to X.Y ?

mechanomi: aha. got you. pushed a fix up

PR #241, opened 2020-09-02
Add integration document for CrateDB & StreamSets

autophagy:
This PR adds an integration document for using CrateDB as a destination in a StreamSet data collector pipeline. This integration document is based on https://crate.io/docs/crate/clients-tools/en/latest/tools/streamsets.html which we wish to remove.

This integration document currently does not have the section about CrateDB as an origin for a pipeline. I will add this after this PR is merged, if we think it is worthwhile.


docs/integrations/streamsets.rst
```diff
@@ -0,0 +1,158 @@
...
+
+This integration document details how to create data streaming pipelines
+using CrateDB and StreamSets Data Collector.
+
+
+Abstract
+========
+
+The `StreamSets Data Collector`_ is a lightweight and powerful engine that
+allows you to build streaming, batch and change data capture pipelines that
```
matthijskrul, line 15: I don't entirely understand this line - are these all nouns compounded to 'data capture pipelines'? or just to 'pipelines'? Are these three things you can build or is building one of three?

docs/integrations/streamsets.rst
```diff
@@ -0,0 +1,158 @@
...
+
+
+Implementation
+==============
+
+
+Set Up
+------
+
+For this implementation, we will ingest and transform the data from
```
matthijskrul, line 32: 2nd person preferably

matthijskrul (approved):
Minor comments

kovrus (approved):

docs/integrations/streamsets.rst
autophagy: Yeah, they all bind to pipelines. There are 3 types:
- Streaming pipelines
- Batch pipelines
- Change-data-capture (CDC) pipelines

What would be a better way of forumulating this?

PR #254, opened 2020-10-12
Remove deb-src entry from crate-CHANNEL.list

mfussenegger:
We don't publish source packages.

Closes https://github.com/crate/crate/issues/10625


mfussenegger:
Just realized that there's also https://github.com/crate/crate-howtos/pull/253 which removes it as well. I guess this here could be closed then.

PR #257, opened 2020-10-20
Trim headlines within navigation/toctree at "Integrations" section

amotl:
Dear @proddata, @mechanomi and @infoverload,

this implements @proddata's suggestions from [1] - thanks for that! It delivers short titles for the navigation on the left hand side but fully verbose titles on the respective pages. The way to do it like that with Sphinx will hopefully make both SEO and our users' eyes more happy.

With kind regards,
Andreas.

P.S.: I didn't scan other parts of the documentation yet, but it can well be expanded to every spot where this makes sense to improve user experience right away without any significant efforts.

P.P.S.: I made this change blindly without running Sphinx on that. Nevertheless, I hope everything will turn out well. @crate/tech-writing: Please adjust wording and stuff if you like, I just took the labels verbatim from the left hand side of the screenshot @proddata shared the other day [2].

[1] https://crate.slack.com/archives/CQ3NEU05A/p1602250372005200
[2] image

### DIFF
--- a/docs/integrations/index.rst
+++ b/docs/integrations/index.rst
@@ -10,22 +10,22 @@ tools.
 .. toctree::
    :maxdepth: 1
 
-   kafka-connect
-   streamsets
+   Apache Kafka <kafka-connect>
+   StreamSets <streamsets>
 
 .. rubric:: Enrichment
 
 .. toctree::
     :maxdepth: 1
 
-    azure-functions
+    Azure IoT Hub and Functions <azure-functions>
 
 .. rubric:: Data Analysis, Visualization, and Machine Learning
 
 .. toctree::
     :maxdepth: 1
 
-    r
-    powerbi-desktop
-    powerbi-gateway
-    ml-dist
+    R <r>
+    PowerBI Desktop <powerbi-desktop>
+    PowerBI Real Time Reports <powerbi-gateway>
+    TensorFlow <ml-dist>


proddata (approved):
lgtm

Though this als changes the content of the "Integrations" page and not only the TOC

amotl:
> This also changes the content of the "Integrations" page and not only the TOC.

Oh, so true. Thanks for spotting this. Nevermind then, let's just close this PR again, right? You decide...

infoverload:
@amotl I like the changed headlines on the left side, but I think the Integrations page should have the full titles.  Is there a way to have both?

mechanomi:
I've been experimenting with this locally today. I thought I could trick Sphinx into doing both but unfortunately I do not think this is possible without coming up with some sort of kludge

I also experimented with shortening the titles to try to drop the use of "CrateDB" on the basis that search engines do already append "CrateDB: How-To Guides" to the page titles on the search result pages. I'm not happy with the result though (see next para)

I think we should change the theme to add a breadcrumb trail above or below each page title. e.g., on the "Data Ingestion using Kafka and Kafka Connect" page, the breadcrumb trail might look like this:

CrateDB Docs > How-To Guides > Integrations

I think that once we have something like this in place, we could justifiably trim down the verbosity of our titles

thoughts?

amotl:
I believe this was just a failed experiment from my end and can safely be closed, right?

norosa:
yup. thanks @amotl

PR #261, opened 2020-11-18
Fix link errors (as reported by sphinx-build -b linkcheck)

mechanomi:
Additionally, drop notice about upgrading from 0.54 to 0.55. This notice links
to the blog, but WordPress returns a forbidden error. This link could be added
to `linkcheck_ignore` in `conf.py`. However, these versions are so old that we
do not even host the changelogs in the docs anymore.

---

for errors originally reported here: https://github.com/crate/tech-writing-domain/issues/336

### DIFF
--- a/docs/best-practices/migrating-from-mysql.rst
+++ b/docs/best-practices/migrating-from-mysql.rst
@@ -74,13 +74,12 @@ when read from the convert script later.
 Converting date/time types
 --------------------------
 
-The standard output for `date/time types in MySQL
-<http://dev.mysql.com/doc/refman/5.0/en/date-and-time-types.html>`_ is a
-string. However, CrateDB uses a ``long`` type to store timestamps (with
-millisecond precision). To prevent problems with dates that have timezone
-information, use MySQL's builtin ``UNIX_TIMESTAMP`` function to convert
-``date``, ``time``, ``datetime``, ``timestamp`` and ``year`` types into UNIX
-timestamps directly in the ``SELECT`` statement.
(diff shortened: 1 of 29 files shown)


docs/best-practices/migrating-from-mysql.rst
```diff
@@ -74,13 +74,12 @@ when read from the convert script later.
...
-string. However, CrateDB uses a ``long`` type to store timestamps (with
-millisecond precision). To prevent problems with dates that have timezone
-information, use MySQL's builtin ``UNIX_TIMESTAMP`` function to convert
-``date``, ``time``, ``datetime``, ``timestamp`` and ``year`` types into UNIX
-timestamps directly in the ``SELECT`` statement.
+The standard output for `date/time types in MySQL`_ is a string. However,
+CrateDB uses a ``long`` type to store timestamps (with millisecond precision).
+To prevent problems with dates that have timezone information, use MySQL's
+builtin ``UNIX_TIMESTAMP`` function to convert ``date``, ``time``,
+``datetime``, ``timestamp`` and ``year`` types into UNIX timestamps directly in
```
matthijskrul, line 81: Last Oxford comma after 'timestamp'?

matthijskrul (approved):

docs/best-practices/migrating-from-mysql.rst
mechanomi: good catch. thx! I'm gonna leave this in as it is unrelated to the changes and it's good to keep things separate (it makes backporting and cherry-picking a lot easier when changes are self-contained)

PR #263, opened 2020-12-17
Correction: Use HTTPS instead of HTTP for links

mechanomi:

see crate/tech-writing-domain#316

### DIFF
--- a/docs/performance/inserts/parallel.rst
+++ b/docs/performance/inserts/parallel.rst
@@ -116,7 +116,7 @@ throughput of your cluster with different setups and under different loads.
 .. _Connection: https://docs.oracle.com/javase/8/docs/api/java/sql/Connection.html
 .. _DataSource: https://docs.oracle.com/javase/8/docs/api/javax/sql/DataSource.html
 .. _Executor: https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/Executor.html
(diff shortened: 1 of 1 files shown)


docs/clustering/multi-node-setup.rst
```diff
@@ -461,8 +461,8 @@ Edit the `transport.tcp.port`_ setting in your `configuration`_ file:
     `More information about port settings`_
```
mkleen, line 461: This section seems broken to me.

mechanomi: @mkleen oh you're totally right. pushed a fixup. thx!

mkleen: Screenshot 2020-12-18 at 14 47 59

mkleen: This is what i meant with broken section.

mechanomi: @mkleen I think this must be an issue with your local dev setup. can you try running `make reset` inside `docs/` and then try again?

you can view the correct output on RTD here:

https://crate-howtos--263.org.readthedocs.build/en/263/clustering/multi-node-setup.html

this (the RTD build) looks fine to me

mkleen: oh you are right, sorry for that.

mkleen (approved):

PR #265, opened 2021-01-21
Correction: Define `POD_NAME` variable used in StatefulSet config

mechanomi:
fixes https://github.com/crate/crate-howtos/issues/264

The StatefulSet configuration tries to set `node.name` using a variable named
`POD_NAME`. However, no such variable is defined. This change defines
`POD_NAME` as a variable that gets its value from the `metadata.name` field.

### DIFF
--- a/docs/deployment/containers/kubernetes.rst
+++ b/docs/deployment/containers/kubernetes.rst
@@ -291,6 +291,10 @@ CrateDB 3.0.5 cluster:
               value: "3"
             - name: CLUSTER_NAME
               value: "my-crate"
+            - name: POD_NAME
+              valueFrom:
+                fieldRef:
+                  fieldPath: metadata.name
             - name: NAMESPACE
               valueFrom:
                 fieldRef:


docs/deployment/containers/kubernetes.rst
```diff
@@ -291,6 +291,10 @@ CrateDB 3.0.5 cluster:
               value: "3"
             - name: CLUSTER_NAME
               value: "my-crate"
+            - name: POD_NAME
+              valueFrom:
+                fieldRef:
+                  fieldPath: metadata.name
```
WalBeh, line 297: Yep, that should work. On the cloud-stack we use .... 

```
        - name: POD_NAME
          valueFrom:
            fieldRef:
              apiVersion: v1
              fieldPath: metadata.name
```

WalBeh (approved):

docs/deployment/containers/kubernetes.rst
mechanomi: @WalBeh is it desirable to include `apiVersion` in the docs example?

PR #266, opened 2021-01-25
Improvement: add warning for partially invalid UNNEST operations

matthijskrul:
- Adds a warning to the Crate Howtos about SELECT FROM UNNEST operations where invalid type combinations may lead to inconsistent results without explicit error handling
- Corresponds (and links) to [reference](https://crate.io/docs/crate/reference/en/4.3/sql/statements/insert.html?highlight=unnest#insert-from-dynamic-queries-constraints)
- Resolves https://github.com/crate/crate-howtos/issues/203 

## Checklist

### DIFF
--- a/docs/performance/inserts/methods.rst
+++ b/docs/performance/inserts/methods.rst
@@ -14,6 +14,7 @@ Choosing the best insert method is an easy way to improve insert performance.
 .. contents::
    :local:
 
+
 .. _insert_statement_types:
 
 Statement types
@@ -28,6 +29,7 @@ The three types of insert statements are:
 Each type of statement can be issued using any of the three types of
 :ref:`client approaches <inserts_client_approaches>`, which we cover in
 the next section.
+
 
 .. _inserts_single_inserts:
 
@@ -62,6 +64,7 @@ insert workloads.
 In addition, lots of internal traffic will congest your network, which will
 slow down any network-based cluster operations (i.e. inserts, distributed
 reads, replication, and cluster management).
+
 
 .. _inserts_unnest:
 
@@ -123,6 +126,16 @@ For example, using the CrateDB Python client, the following is possible:
 Here, you can vary the number of rows being inserted without having to change
 the prepared statement.
 
+.. WARNING::
+
+    When inserting using ``UNNEST``, CrateDB may drop rows that produce errors
+    without returning an error message. This happens when the ``SELECT`` using
+    ``UNNEST`` affects rows with invalid column names, or with data types that
+    are not internally consistent. This behavior can produce inconsistencies
+    and unexpected results. Refer to the `UNNEST reference documentation`_ for
+    more detail.
+
+
 .. _inserts_multiple_values:
 
 Multiple value expressions
@@ -150,6 +163,7 @@ Query string parameter substitution is recommended over string concatenation,
 and so the ``UNNEST`` method is recommended over the multiple value expressions
 method.
 
+
 .. _inserts_client_approaches:
 
 Client approaches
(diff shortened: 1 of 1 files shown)


docs/performance/inserts/methods.rst
```diff
@@ -123,6 +126,15 @@ For example, using the CrateDB Python client, the following is possible:
 Here, you can vary the number of rows being inserted without having to change
 the prepared statement.
 
+.. warning::
```
mechanomi, line 129:
```suggestion
.. WARNING::
```

mechanomi: I would also add an empty newline after this

docs/performance/inserts/methods.rst
```diff
@@ -123,6 +126,15 @@ For example, using the CrateDB Python client, the following is possible:
...
+    When inserting using ``UNNEST``, e.g. using the Python client, rows that
```
mechanomi, line 130: I would put the e.g. inside parentheses here

I think this sentence could also benefit from being rewritten to use the active voice

docs/performance/inserts/methods.rst
```diff
@@ -123,6 +126,15 @@ For example, using the CrateDB Python client, the following is possible:
...
+    produce errors while inserting may be dropped without an error message
+    being returned. This can produce inconsistencies and unexpected results.
```
mechanomi, line 132: indeterminate reference. perhaps "This behavior" or similar

docs/performance/inserts/methods.rst
```diff
@@ -123,6 +126,15 @@ For example, using the CrateDB Python client, the following is possible:
...
+    This happens when the ``SELECT`` using ``UNNEST`` involves rows with
```
mechanomi, line 133: perhaps this example/explanation should be moved above the concluding statement about the behavior producing inconsistencies (to improve narrative flow)?

docs/performance/inserts/methods.rst
matthijskrul: I'd agree, except for the 'drop' verb: that refers to unintended behavior, which feels more natural put in passive tense. Alternatively you'd get "rows... may drop", which sounds odder to me.

mechanomi: in this case, the doer is CrateDB, so I would imagine it more like "CrateDB may drop rows when...". what do you think?

matthijskrul: That's a good idea!

mechanomi: just looking at this again and reviewing the note in context. I think we can drop the bit about the "Python client". the behavior being described is not specific to the Python client. the only bit which is specific to the Python client is the code example above but I don't think that matters for this warning

docs/performance/inserts/methods.rst
```diff
@@ -123,6 +126,16 @@ For example, using the CrateDB Python client, the following is possible:
 Here, you can vary the number of rows being inserted without having to change
 the prepared statement.
 
+.. WARNING::
+
+    When inserting using ``UNNEST``, (e.g.) using the Python client, CrateDB
```
mechanomi, line 131:
```suggestion
    When inserting using ``UNNEST``, CrateDB
```

will need re-wrapping

matthijskrul: Or "e.g. when you use..."? Assuming that's what you mean by wrapping

mechanomi: I think we can just drop the whole bit about "when using the Python client". this advice applies to any client. at least, that's my understanding! (only caught this when reviewing again today sorry)

matthijskrul: Alright! It was there in the original so I didn't want to fiddle with it

mechanomi (approved):

PR #268, opened 2021-01-25
Improvement: Switch from `/srv/crate` to `/var/lib/crate` for external volumes

mechanomi:
(closes https://github.com/crate/crate-howtos/issues/211)

Also included: general copyedit for this subsection.

### DIFF
--- a/docs/going-into-production.rst
+++ b/docs/going-into-production.rst
@@ -317,22 +317,39 @@ directories off to a persistent location. You can do this using the
 
     `Path settings`_
 
-For example, if you are running CrateDB on a `Unix-like`_ operating system, the
-`Filesystem Hierarchy Standard`_ (FHS) recommends the ``/srv`` directory as the
-root for site-specific data.
+If you are following the `shared-nothing`_ approach to deployment, the best way
+to handle persistent data is to keep it on an external volume. This allows you
+to persist data beyond the lifespan of an individual virtual machine or
+container.
 
-With this in mind, if you are installing CrateDB by hand, a good value for
-`CRATE_HOME`_ on a Unix-like system might be ``/srv/crate``. Make sure to set
-`CRATE_HOME` before running `bin/crate`_.
+.. CAUTION::
+
+    This is required if you are using Docker, which is stateless by design.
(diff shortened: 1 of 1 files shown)


justien (commented):
lgtm!

amotl (changes requested):
Hi Naomi,

I would not recommend `/mnt` for production use. `/mnt` is meant to be a temporary mount point used by system administrators, see also [1,3]. Also, `/media` would not fit as it designates a mount point used for removable devices [2,3]. I also wonder why `/srv` was chosen here beforehand.

I would strongly suggest to follow the principle of least surprise and recommend to use `/var/lib/cratedb` as other database systems are doing so. For example, `/var/lib/mysql`, `/var/lib/mongodb`, `/var/lib/influxdb` and so forth. This is also clearly reflected within the FHS [4]:

> Directory | Description
> -- | --
> /srv | Site-specific data served by this system, such as data and scripts for web servers, data offered by FTP servers, and repositories for version control systems (appeared in FHS-2.3 in 2004).
> /var/lib | State information. Persistent data modified by programs as they run, e.g., databases, packaging system metadata, etc.

<br/>

With kind regards,
Andreas.


[1] https://tldp.org/LDP/Linux-Filesystem-Hierarchy/html/mnt.html
[2] https://tldp.org/LDP/Linux-Filesystem-Hierarchy/html/media.html
[3] https://unix.stackexchange.com/questions/13975/mounting-a-device-role-of-dev-media-and-mnt-and-the-mount-command
[4] https://en.wikipedia.org/wiki/Filesystem_Hierarchy_Standard#Directory_structure

mechanomi:
@amotl omg thanks! very clear. makes sense. pushed a fixup

amotl:
Thanks for fixing this. In order to discriminate between Rust's Cargo package management system [1], which is using the term "Crate" for its artifacts, I would really like to see `/var/lib/cratedb` being used here. However, our distribution packages already nailed it to use `/var/lib/crate` [2], so we should probably keep it like that.

[1] https://doc.rust-lang.org/cargo/
[2] https://github.com/crate/distribute

amotl (approved):

PR #270, opened 2021-02-08
Improve recommendation about using an external volume for data

mechanomi:
(closes https://github.com/crate/crate-howtos/issues/211)

Specifically:

- Switch to `/opt/cratedb` per the discussion on
  https://github.com/crate/crate-howtos/issues/211

- Note that it is not necessary to configure some data paths if you have set
  `CRATE_HOME` and are happy with the defaults

- Explain why the use of `/opt/cratedb` differs from the default directories
  used by other installation methods


amotl (approved):
Thank you!

PR #274, opened 2021-03-24
Bulk inserts overhaul plus additional fixes and improvements

norosa:
this PR does the following:

- upgrades the common build system to use crate-docs 1.0

- merges the "importing huge data sets" content into the "inserts performance"

- while I was at it, I decided to give this whole document an overhaul (see the commit message for more info). this was split into a separate commit from the move so the diff can be read easily

- fixes broken links now reported by the build

- fixes Vale errors now reported by the build


docs/performance/inserts/bulk.rst
```diff
@@ -0,0 +1,419 @@
+.. highlight:: psql
+
+.. _bulk-inserts:
+
+============
+Bulk inserts
+============
+
+If you have a lot of pre-existing data that you need to import in bulk, follow
+this how-to guide if you want to improve performance.
```
amotl, line 10:
```suggestion
this how-to guide to get optimal performance on this task.
```

docs/performance/inserts/bulk.rst
```diff
@@ -0,0 +1,419 @@
...
+
+Shards and replicas
+-------------------
+
+For each table, if you do not set the number of `shards`_ or the number of
+`replicas`_, the default configuration is as follows:
+
+**Shards**
+    Dependant on the number of data-nodes in the cluster (see `CLUSTERED
+    clause`_)
```
amotl, line 97: Not sure about this?
```suggestion
    Dependent on the number of data-nodes in the cluster (see `CLUSTERED
    clause`_).
```

docs/performance/inserts/bulk.rst
```diff
@@ -0,0 +1,419 @@
...
+**Replicas**
+    One
+
+You should choose the number of shards wisely. The choice depends on the number
+of nodes in the cluster as well as on the amount of data that is going into the
+table.
+
+Assuming there are six nodes in the cluster, you can put two shards on each
+node, giving you a total of 12 shards. This should be enough to handle millions
+of records.
```
amotl, line 108:
```suggestion
node, giving you a total of 12 shards. This should be enough to store millions
of records efficiently.
```

docs/performance/inserts/bulk.rst
```diff
@@ -0,0 +1,419 @@
...
+
+Shards can be thought of as "virtual nodes." You should create enough for your
```
amotl, line 110:
```suggestion
Shards can be thought of as "virtual nodes". You should create enough for your
```

docs/performance/inserts/bulk.rst
```diff
@@ -0,0 +1,419 @@
...
+----------------
+
+Another way to speed up importing is to set the `refresh_interval`_ of the
+table to zero::
+
+    cr> ALTER TABLE users SET (refresh_interval = 0);
+    ALTER OK, -1 rows affected (... sec)
+
+This will disable the periodic `refresh`_ of the table that is needed to
+minimize the effect of eventual consistency and also minimize the overhead
```
amotl, line 148:
```suggestion
minimize the effect of eventual consistency in order to minimize the overhead
```

docs/performance/inserts/bulk.rst
```diff
@@ -0,0 +1,419 @@
...
+
+.. _bulk-import-data:
+
+Import the data
+===============
+
+Once the table is created, you can start importing the data.
+
+When importing, CrateDB has `native support for JSON data`_, where each line is
+a JSON string representing a single record. Empty lines are skipped. The keys
```
amotl, line 190: This format is called "JSON Lines" [1], in order to discriminate against regular JSON. Can you add this reference here in a good way?

> JSON Lines files may be saved with the file extension `.jsonl`.

[1] https://jsonlines.org/

docs/performance/inserts/bulk.rst
```diff
@@ -0,0 +1,419 @@
...
+of the JSON objects are mapped to columns when imported. Nonexistent columns
+will be created if necessary.
+
+For example, a JSON file might look like this:
```
amotl, line 194:
```suggestion
For example, a `.jsonl` file might look like this:
```

amotl, line 194:
```suggestion
For example, a ``.jsonl`` file might look like this:
```

amotl, line 194: For purposes of correctness, I would prefer this. However, if CrateDB does not grok importing from `.jsonl` files, i.e. if it does not consider it as a valid file extension, please just skip this suggestion.
```suggestion
For example, a ``.jsonl`` file might look like this:
```

docs/performance/inserts/bulk.rst
```diff
@@ -0,0 +1,419 @@
...
+    {"id": 1, "name": "foo", "day_joined": 1408312800, "bio": "Lorem ipsum dolor sit amet, consectetuer adipiscing elit.", "address": {"city": "Dornbirn", "country": "Austria"}}
+    {"id": 2, "name": "bar", "day_joined": 1408312800, "bio": "Lorem ipsum dolor sit amet, consectetuer adipiscing elit.", "address": {"city": "Berlin", "country": "Germany"}}
+
+
+Use the `COPY FROM`_ statement to import JSON data directly from a file::
+
+    cr> COPY users FROM '/tmp/best_practice_data/users.json';
+    COPY OK, 150 rows affected (... sec)
+
+Here, CrateDB will check ``/tmp/best_practices_data/`` locally on each node in
```
amotl, line 207: I believe there have to be special considerations when referencing files on Windows file systems. Maybe it would be appropriate to add them below this section as an appropriate `.. TIP` admonition?

docs/performance/inserts/bulk.rst
```diff
@@ -0,0 +1,419 @@
...
+    component of the primary key.
+
+A value identifying the target partition column *should* be defined in the
+``COPY FROM`` statement using the ``PARTITION`` clause::
+
+    cr> COPY users PARTITION (day_joined=1408312800)
+    ... FROM '/tmp/best_practice_data/users_1408312800.json';
+    COPY OK, 23 rows affected (... sec)
+
+When you use the ``PARTITION`` clause, CrateDB doesn't need to resolve the
```
amotl, line 335:
```suggestion
When using the ``PARTITION`` clause, CrateDB doesn't need to resolve the
```

docs/performance/inserts/bulk.rst
```diff
@@ -0,0 +1,419 @@
...
+partition for each row that is imported, which aids performance.
```
amotl, line 336:
```suggestion
partition for each row that is imported, which also improves performance.
```

docs/performance/inserts/bulk.rst
```diff
@@ -0,0 +1,419 @@
...
+    :ref:`Bulk inserts: Shards and replicas <bulk-shards-replicas>`
+
+
+.. _bulk-increase-shards:
+
+Increase the number of shards
+-----------------------------
+
+The ``ALTER TABLE`` clause can also be used to `alter the number of shards`_
+for newly created partitions, which may improve improve performance when
```
amotl, line 385:
```suggestion
for newly created partitions, which may improve performance when
```

docs/performance/inserts/bulk.rst
```diff
@@ -0,0 +1,419 @@
...
+handling a lot more data than before.
```
amotl, line 386:
```suggestion
handling a lot more data than before, when the table was initially created
with a specific number of shards.
```

amotl (commented):
Thank you very much for improving this part of the documentation. I added some suggestions which came to my mind. As always, take the ones you like and skip the others which don't seem justified.

docs/build.json
```diff
@@ -1,5 +1,5 @@
 {
   "schemaVersion": 1,
   "label": "docs build",
-  "message": "0.4.0"
+  "message": "1.0.0"
```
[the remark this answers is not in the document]
norosa: I only just released it but yeah let's haha

docs/Makefile
```diff
@@ -18,45 +18,128 @@
...
+
+# Skip if no version could be determined (i.e., because of a network error)
+ifneq ($(LATEST_VERSION),)
+# Only issue a warning if there is a version mismatch
+ifeq ($(BUILD_VERSION),$(LATEST_VERSION))
+define version_warning
+You are using Crate Docs version $(BUILD_VERSION), however version \
+$(LATEST_VERSION) is available. You should consider upgrading. Follow the \
+instructions in `Makefile` and then run `make reset`.
+endef
```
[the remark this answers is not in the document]
norosa: this is upstream also. what is the typo?

docs/performance/inserts/bulk.rst
norosa: I think I like my phrasing better, for conciseness

docs/performance/inserts/bulk.rst
norosa: I think the less-specific "handle" here is preferable because sharding is not just about storage capacity, it's also about redundancy and performance

docs/performance/inserts/bulk.rst
norosa: this is incorrect. punctuation should be inside the quotes

https://www.grammarly.com/blog/quotation-marks/#:~:text=Do%20commas%20and%20periods%20go,go%20inside%2C%20sometimes%20stay%20outside.

docs/performance/inserts/bulk.rst
norosa: good catch. reworded slightly

docs/performance/inserts/bulk.rst
norosa: opted to say "JSON Lines" hereo

norosa: as above

docs/performance/inserts/bulk.rst
norosa: disagree. the emphasis here is on the choice the user makes whether to use it or not use it

docs/performance/inserts/bulk.rst
norosa: disagree. the "also" is incorrect because the only reason we're mentioning the previous clause is *because* it has an impact on performance

additionally, I considered using "improves" here, but I shied away from it because I think it's too-definitive a statement to make. "may improve performance", on the other hand, sounds too meek

I settled on "aids" because it strikes the middle ground of communicating that it "helps" performance even if it practically makes no difference

I didn't use "helps" because I don't feel like "which helps performance" strictly good grammar in this context because there is a verb missing (i.e., helps performance do WHAT?)

docs/performance/inserts/bulk.rst
norosa: slightly reworded

norosa:
@amotl thanks so much for your characteristically comprehensive review!!

I dismissed a few of your suggestions with what I hope are reasonable explanations why

I tried to upgrade to 2.0, but there is some sort of bug in the Makefile. ugh. I am keeping it pinned at 1.0 for now, which at least works

can you review again please? ty ty!

docs/performance/inserts/bulk.rst
norosa: used "JSON Line"

docs/Makefile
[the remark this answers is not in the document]
amotl: I wasn't aware that `endef` is actually a thing, but learned from [1] now. Apologies!

[1] https://www.gnu.org/software/make/manual/html_node/Multi_002dLine.html

docs/performance/inserts/bulk.rst
amotl: Thanks for improving my English writing along the lines and sorry for the noise.

amotl (approved):
Thank you very much again for incorporating the minor fixes, where appropriate, as well as the mentioning of the "JSON Lines" file format and the addition of the TIP for Windows users when importing JSON Lines files.

Hereby, I am approving this update. However, maybe @proddata might want to also review it (maybe also in retrospective), because he is the one who is more in touch with users, their requirements and perspectives. Saying that, I believe he might also be able to review this on behalf of the final (new) document and add eventual refinements within another iteration.

proddata (approved):

PR #278, opened 2021-04-09
Fix broken links

amotl:
The new nightly link checker job works like a charm, see [1].

[1] https://github.com/crate/crate-howtos/runs/2303702717


proddata (commented):
👍

PR #294, opened 2021-09-15
Mention a lower boundary for memory configuration

mfussenegger:
Although it is possible to start up an instance with 512MB or even something as low as 256MB, we typically don't test with values that low, so we should recommend at least 1GB


jweberpm (approved):
Looks good to me.

seut (approved):

PR #300, opened 2022-02-21
Use headless service for node discovery

proddata:
## Checklist

### DIFF
--- a/docs/deployment/containers/kubernetes.rst
+++ b/docs/deployment/containers/kubernetes.rst
@@ -129,9 +129,9 @@ Here's an example configuration snippet:
       labels:
         app: crate
     spec:
-      # A static IP address is assigned to this service. This IP address is
-      # only reachable from within the Kubernetes cluster.
-      type: ClusterIP
+      # A Headless Service is used for CrateDB node discovery.
+      # Returns the IP adresses of the associated CrateDB Pods.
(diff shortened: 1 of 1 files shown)


docs/deployment/containers/kubernetes.rst
```diff
@@ -129,9 +129,9 @@ Here's an example configuration snippet:
       labels:
         app: crate
     spec:
-      # A static IP address is assigned to this service. This IP address is
-      # only reachable from within the Kubernetes cluster.
-      type: ClusterIP
```
SStorm, line 134: You will still need the type. What makes a headless service is the `clusterIP: None` thing.

proddata: ClusterIP is the default type.
In most definition I found it was omitted for headless services.
The comment is definitely wrong

SStorm: Ah yes you're right, just tried it and it works fine w/o `type`

SStorm (approved):

protosam:
I think the changes should look more like this: https://github.com/protosam/crate-howtos/blob/master/docs/deployment/containers/kubernetes.rst

I would just make a PR, but the contributor's agreement needs amendments before I could agree to it. Will discuss that in the community discuss though.

PR #307, opened 2022-07-13
rolling upgrades from prev latest minor to the next major are allowed

BaurzhanSakhariev:


mfussenegger (approved):
lgtm, but I wonder if it could do with some rephrasing to make it a bit shorter. E.g. instead of stating the special cases, just mention the supported case?

E.g. "A rolling upgrade is possible from one version to the next feature version. Some examples:

- You can do a rolling upgrade from 4.5.x to 4.6.0
- You can do a rolling upgrade from 4.8.x to 5.0.0 (because 4.8 is the last feature release within the `4.` series.
- You cannot do a rolling do a rolling release from `x.y.z` to `x.(y + 3).z` unless the release notes explicitly mention support.

BaurzhanSakhariev:
> lgtm, but I wonder if it could do with some rephrasing to make it a bit shorter. E.g. instead of stating the special cases, just mention the supported case?

Thanks for the suggestion, I added examples and removed patch/ new minor things completely from the WARNING - as smaller upgrades possibility implies from bigger upgrade. Only special cases like testing/stable versions are kept in the warning.

PR #314, opened 2022-12-29
Clarifications around metadata gateway settings

hlcianfagna:
Improving the documentation section about metadata gateway settings, explaining and emphasizing a few details.

### DIFF
--- a/docs/clustering/multi-node-setup.rst
+++ b/docs/clustering/multi-node-setup.rst
@@ -95,11 +95,12 @@ instructions.
 
        sh$ tar -xzf crate-*.tar.gz
 
-2. You should configure the :ref:`metadata gateway <metadata-gateway>` so that
-   CrateDB knows how to recover its state safely. Ideally, for a three-node
-   cluster, set `gateway.expected_data_nodes`_ to **3** and set
-   `gateway.recover_after_data_nodes`_ to **3**. You can specify these settings
-   in the `configuration`_ file of the unpacked directory.
+2. It is common to configure the :ref:`metadata gateway <metadata-gateway>` so 
+   that the cluster waits for all data nodes to be online before starting the
+   recovery of the shards. In this case let's set 
+   `gateway.expected_data_nodes`_ to **3** and 
+   `gateway.recover_after_data_nodes`_ also to **3**. You can specify these 
+   settings in the `configuration`_ file of the unpacked directory.
 
    .. NOTE::
 
@@ -162,9 +163,9 @@ To run a CrateDB cluster across multiple hosts, you must manually configure the
 bootstrapping process by telling nodes how to:
 
   a. :ref:`Discover other nodes <node-discovery>`
-  b. :ref:`Elect a master node <master-node-election>`
+  b. :ref:`Elect a master node <master-node-election>` the first time
 
-You must also configure the :ref:`metadata gateway <metadata-gateway>` (as with
+You can also configure the :ref:`metadata gateway <metadata-gateway>` (as with
 auto-bootstrapping).
 
 
@@ -314,29 +315,26 @@ ensures that the cluster does not elect multiple masters in the event of a
 network partition (also known as a `split-brain`_ scenario).
 
 CrateDB (versions 4.x and above) will automatically determine the ideal `quorum
-size`_. If you are using CrateDB versions 3.x and below, you must manually set
-the quorum size using the `discovery.zen.minimum_master_nodes`_ setting.
-
-.. NOTE::
-
-    For a three-node cluster, CrateDB will set the quorum size to three.
-    Consequentially, you must declare all nodes to be master-eligible. Consult
-    the `quorum guide`_ for detailed information about quorum size
-    considerations.
-
-If you configure fewer master-eligible nodes than the ideal quorum size,
-CrateDB will issue a warning (visible in the logs and the `admin UI`_).
-
+size`_, but if you are using CrateDB versions 3.x and below, you must manually set
+the quorum size using the `discovery.zen.minimum_master_nodes`_ setting and for 
+a three-node cluster, you must declare all nodes to be master-eligible. 
 
 .. _metadata-gateway:
 
 Metadata gateway
 ^^^^^^^^^^^^^^^^
 
-When running a multi-node cluster, you must configure the `metadata gateway`_
-so that CrateDB knows how to recover its state. For a three-node cluster, set
-`gateway.expected_data_nodes`_ to **3** and `gateway.recover_after_data_nodes`_
-to **3**.
+When running a multi-node cluster, you can configure the `metadata gateway`_
+settings so that CrateDB delays recovery until a certain number of nodes is
+available.
+This is useful because if recovery is started when some nodes are down 
+CrateDB will proceed on the basis the nodes that are down may not be coming 
+back, and it will create new replicas and rebalance shards as necessary. 
+This is an expensive operation that, depending on the context, may be better 
+avoided if the nodes are only down for a short period of time.
+So, for instance, for a three-nodes cluster, you can decide to set
+`gateway.expected_data_nodes`_ to **3**, and 
+`gateway.recover_after_data_nodes`_ also to **3**.
 
 You can specify both settings in your `configuration`_ file:
 
@@ -346,7 +344,7 @@ You can specify both settings in your `configuration`_ file:
       recover_after_data_nodes: 3
       expected_data_nodes: 3
 
-Alternatively, you can configure this setting at startup with command-line
+Alternatively, you can configure these settings at startup with command-line
 options:
 
 .. code-block:: console
@@ -450,7 +448,7 @@ Edit the `transport.tcp.port`_ setting in your `configuration`_ file:
 
 .. TIP::
 
-    If you are running a node on Docker, you must configure CrateDB to publish the
+    If you are running a node on Docker, you can configure CrateDB to publish the
     container's external hostname and the external port number bound to the
     transport port. You can do that in your `configuration`_ file using the
     `network.publish_host`_ and `transport.publish_port`_ settings.


docs/clustering/multi-node-setup.rst
```diff
@@ -319,10 +319,14 @@ the quorum size using the `discovery.zen.minimum_master_nodes`_ setting.
 
 .. NOTE::
 
-    For a three-node cluster, CrateDB will set the quorum size to three.
-    Consequentially, you must declare all nodes to be master-eligible. Consult
-    the `quorum guide`_ for detailed information about quorum size
-    considerations.
+    CrateDB will set the quorum size to `FLOOR(n / 2) + 1` so that is the
```
amotl, line 322:
```suggestion
    CrateDB will set the quorum size to ``FLOOR(n / 2) + 1``, so that is the 
```

amotl (changes requested):

docs/clustering/multi-node-setup.rst
```diff
@@ -319,10 +319,14 @@ the quorum size using the `discovery.zen.minimum_master_nodes`_ setting.
 
 .. NOTE::
 
-    For a three-node cluster, CrateDB will set the quorum size to three.
-    Consequentially, you must declare all nodes to be master-eligible. Consult
-    the `quorum guide`_ for detailed information about quorum size
-    considerations.
+    CrateDB will set the quorum size to ``FLOOR(n / 2) + 1``, so that is the 
+    minimum number of master-eligible nodes you should have.
+    For high availability purposes you want to have one additional
```
hammerhead, line 324: > For high availability purposes you want to have one additional master-eligible node on top of the minimum required for quorum for each node loss you want to tolerate.

It might be just me, but I was having trouble understanding this part until I re-read it several times.
How about something like:
> For high availability, increase the number of master-eligible nodes by one for each node loss you want to compensate.

docs/clustering/multi-node-setup.rst
```diff
@@ -319,10 +319,14 @@ the quorum size using the `discovery.zen.minimum_master_nodes`_ setting.
...
+    master-eligible node on top of the minimum required for quorum for each
+    node loss you want to tolerate. A 3-nodes cluster for instance with
```
hammerhead, line 326: For consistency, further down in the file the spelling "three-node cluster" is used.

docs/clustering/multi-node-setup.rst
hlcianfagna: Doing further tests I found that the requirement for a quorum of data nodes does not seem to extend to the number of master-eligible nodes, it seems sufficient to have one master-eligible node among the nodes that are available. No warning seems to be triggered when the number of master eligible nodes is below "quorum level", so I removed the entire section from this document.

hlcianfagna: Looking further I believe nodes that are not master-eligible still carry a copy of the cluster state, so at that point I think it makes sense to only check for a quorum of data-carrying nodes.

seut: Yes right, the quorum part is only relevant for CrateDB < 4.x. Just to clarify and not relevant to this change (but more to your comment), on a fresh new cluster, `initial_master_nodes` must be set to > 1 as these are the nodes which are allowed to vote at the first ever master node election. If only 1 node would be listed here, election wouldn't work. But as noted this is only needed on a fresh cluster (and in production), after the first election has taken place, this setting is not needed anymore and is even recommended to be removed.

docs/clustering/multi-node-setup.rst
```diff
@@ -309,34 +309,16 @@ For example:
...
 ^^^^^^^^^^^^^^^^
 
 When running a multi-node cluster, you must configure the `metadata gateway`_
-so that CrateDB knows how to recover its state. For a three-node cluster, set
-`gateway.expected_data_nodes`_ to **3** and `gateway.recover_after_data_nodes`_
-to **3**.
+so that CrateDB knows how to recover its state. For a three-nodes cluster, set
+`gateway.expected_data_nodes`_ to **3**, `gateway.recover_after_data_nodes`_
+could be either **2** or **3**, only set it to **2** if you plan to have at
+least 1 replica for all your tables.
```
seut, line 321: I'm not sure if we should document the `2` variant as it may only increase complexity and confusion like: why only 2 if I have replicas configured? It's valid but not sure how much it helps specially for non super experts.

docs/clustering/multi-node-setup.rst
hlcianfagna: Thank you.
My thinking removing the section in the latest revision in this branch is to make the how-to more readable for people looking at new deployments which would be with versions >=4.0.0

Regarding `initial_master_nodes` listing only 1 node, for new 5.1.2 clusters, I have confirmed this works both 
- for a 3 nodes cluster (the first time the node that is listed gets elected as master, but in subsequent cold starts another node gets elected without issues if the node that was listed for the first election is not available)
- and for a 1-node cluster (`discovery.seed_hosts` set to the hostname of the node - `recover_after_data_nodes`=1 - `expected_data_nodes`=1), later scaled to 2 and then 3 nodes. Tried both individual node restarts and full stop and start of just 2 nodes and master election seems to work fine.

seut: Did you try that out on nodes not bound the loopback interface? If bound to the loopback interface the cluster is treat as in-development and thus master election will work even without any `initial_master_nodes` set.

docs/clustering/multi-node-setup.rst
hlcianfagna: I just tested with a table with `number_of_replicas`=0 , in an environment with `expected_data_nodes`=3 and `recover_after_data_nodes`=2. Starting then with 2 nodes, tables for which shards are available work normally, and the table missing shards does not work, but it resumes working as normal once the 3rd node joins back.
Do you know of some circumstances where this would not be safe? if it is safe maybe  the how to should point to the 2 variant only? what I want to pre-empt here is the somehow counterintuitive situation where a 3-nodes cluster, that can withstand without issues the shutdown of one node, fails to start from cold with the same 2 nodes because the current version of the docs recommends setting `recover_after_data_nodes` to match the expected number of nodes.

docs/clustering/multi-node-setup.rst
seut: As said, this is only related to the very first master election of a fresh cluster. It is irrelevant for nodes joining an existing cluster or node restarts.

docs/clustering/multi-node-setup.rst
seut: And also when replica's are configured. Because of the replica's, all shards for a working table are available, but the replica shards will turn then into primary shards, the missing original primary and replica shards are treat as gone and maybe new replica shards will be created on different nodes. This will result in network and disk load which one want to avoid if the other nodes are just being restart and are coming back soon.

hlcianfagna: >  it will result in a broken table (RED health) as long as the missing shards aren't available

Yes, but this could be acceptable because we get an error message when trying to query the data that is not available, what I worry about is if there are situations where data may get corrupted or changes lost. Could we for instance risk losing changes made since the last refresh interval plus flush interval?

> in rolling-upgrade or any other node restart scenario, you want the cluster to start the recovery of all tables only if all shards are available

I was focusing on a start from cold scenario (imagine power cut - storage controller fried on one node - I still have all nodes except 1 and my tables have replicas - I need to restore service)
For rolling restarts, I think high availability would still be very desiderable, I think it is one of the main reasons why people go for cluster deployments instead of standalone instances.

> the missing original primary and replica shards are treat as gone and maybe new replica shards will be created on different nodes. This will result in network and disk load which one want to avoid if the other nodes are just being restarted

This is a very good point, I need to add a note about this, it came out in a conversation we had with @proddata  2 weeks ago.
For a start from cold, where some nodes come back online faster than other nodes we have `recover_after_time`, but from my tests I am under the impression this does not help in the rolling upgrade scenario, maybe I am missing something or there is some other setting?

proddata: > but from my tests I am under the impression this does not help in the rolling upgrade scenario, maybe I am missing something or there is some other setting?

Still not quite understand what the "issue" with rolling upgrades should be? We even refer in our docs to prevent allocations, by only allowing `new_primaries` https://crate.io/docs/crate/howtos/en/latest/admin/rolling-upgrade.html#step-1-disable-allocations which is btw done automatically for cloud / operator managed clusters.

i.e. during a rolling update, nothing unintended would happen.

seut: Just think of simple rolling restarts instead of rolling upgrades without using graceful stops or disabled allocations.

hlcianfagna: > Could we for instance risk losing changes made since the last refresh interval plus flush interval?

I tested with `"translog.flush_threshold_size"= 100000000,"refresh_interval" = 10000` (durability set to the default which is `REQUEST`) and no data seems to go lost, when there are replicas records inserted just before the abrupt shutdown are available when nodes with replicas come back online and in the case of no replicas the record are available as soon as the node with the replica joins.

proddata: I think we are a mixing up many things in this conversation. Recovery and the related settings are **somewhat** independent from replica settings and therefore I would not at such a note to the documentation. 

In fact 2 data nodes could be perfectly fine in a setup of 2 data nodes and 1 master node, with no replicas. What is "right" largely depends on the setup and expectations. The recovery process also is not "free" and might take some time, why one would want to start recovery, even thought not all nodes are available yet.

proddata: > durability set to the default which is REQUEST

Which ensures that everything gets written to the wal/translog of all shards.

> Just think of simple rolling restarts instead of rolling upgrades without using graceful stops or disabled allocations.

Which is problematic in that sense of that one doesn't know if a write was successful and CrateDB basically auto-commits transactions.

In terms of recovery starting after one node is gone all of a sudden: I don't think it is a big issue. If a node leaves unexpectedly the time how long it takes to come back is most likely also unknown most. Recovery is also delayed by `unassigned.node_left.delayed_timeout` on a table level which defaults to `1m`. So if a node comes back after 2 minutes, recover might have been started, but gets interrupted and the shards on the node get started again.
One could argue for a cluster wide setting instead for comfort reasons.

docs/clustering/multi-node-setup.rst
proddata: We are mixing again, high availability (which depends on the replication settings) and general availability (which depends on a quorum in the clusters voting configuration).

and seemingly also `initial_master_nodes` and `node.master: true`

>  CrateDB will set the quorum size to ``FLOOR(n / 2) + 1```

with `n` being the number of nodes in the voting configuration (!) not the total number of master eligible nodes.
`FLOOR(4/2) + 1 == 3` for a 4 node cluster the quorum however is 2.

The voting config is calculated with:
```java
size - (size % 2 == 0 ? 1 : 0);
```

3 master eligible nodes -> 3 nodes in voting config -> quorum reached with 2
4 master eligible nodes -> 3 nodes in voting config -> quorum reached with 2
5 master eligible nodes -> 5 nodes in voting config -> quorum reached with 3
5 master eligible nodes -> 6 nodes in voting config -> quorum reached with 4

---

> 5-nodes cluster can tolerate the loss of 2 nodes and so on

The cluster itself can tolerate the **simultaneous** loss of 2 nodes.
However you can perfectly fine remove up to 3 nodes from a 5 node cluster for it still to work (as cluster), which does not imply, that all data is available. You just need to wait till the voting configurations adapt to the new number of nodes.

> Doing further tests I found that the requirement for a quorum of data nodes does not seem to extend to the number of master-eligible nodes, it seems sufficient to have one master-eligible node among the nodes that are available

Could you be more explicit what you identified here and maybe post the relevant settings.

docs/clustering/multi-node-setup.rst
```diff
@@ -314,29 +315,26 @@ ensures that the cluster does not elect multiple masters in the event of a
...
-so that CrateDB knows how to recover its state. For a three-node cluster, set
-`gateway.expected_data_nodes`_ to **3** and `gateway.recover_after_data_nodes`_
-to **3**.
+When running a multi-node cluster, you can configure the `metadata gateway`_
+settings so that CrateDB delays recovery until a certain number of nodes is
+available.
+This is useful because if recovery is started when some nodes are down 
+CrateDB will proceed on the basis the nodes that are down may not be coming 
+back, and it will create new replicas and rebalance shards as necessary, 
+this is an expensive operation that, depending on the context, may be better
```
amotl, line 333: Spotted a little chance for improving the reading flow.
```suggestion
back, and it will create new replicas and rebalance shards as necessary. 
This is an expensive operation that, depending on the context, may be better 
```

amotl (changes requested):
Given that a few people contributed to reviewing this documentation improvement already (thanks a stack!), I think it will be safe to be merged? Do you have any other plans, @hlcianfagna?

hlcianfagna:
> Given that a few people contributed to reviewing this documentation improvement already (thanks a stack!), I think it will be safe to be merged? Do you have any other plans, @hlcianfagna?

Hi, I wanted to wait to get feedback on https://github.com/crate/crate/pull/13567 first

docs/clustering/multi-node-setup.rst
```diff
@@ -95,11 +95,12 @@ instructions.
...
-2. You should configure the :ref:`metadata gateway <metadata-gateway>` so that
-   CrateDB knows how to recover its state safely. Ideally, for a three-node
-   cluster, set `gateway.expected_data_nodes`_ to **3** and set
-   `gateway.recover_after_data_nodes`_ to **3**. You can specify these settings
-   in the `configuration`_ file of the unpacked directory.
+2. It is common to configure the :ref:`metadata gateway <metadata-gateway>` so 
+   that the cluster waits for all data nodes to be online before starting the
+   recovery of the shards. In this case let's set 
+   `gateway.expected_data_nodes`_ to **3** and 
+   `gateway.recover_after_data_nodes`_ also to **3**. You can specify these
```
proddata, line 102: Just fyi: in the case of crate-operator and therefore also Cloud `recover_after_data_nodes` is set to `floor(number_nodes / 2) + 1`

proddata (approved):
I left a comment, but otherwise fine 👍

amotl (approved):
Thanks!

PR #318, opened 2022-12-30
Remove unused mergify

hammerhead:
There is a deprecation warning from mergify, and we don't use it anyway in `crate-howtos`/`crate-tutorials`, do we?
Screenshot 2022-12-30 at 15 07 35

## Checklist

 - [X] [CLA](https://crate.io/community/contribute/cla/) is signed


seut:
I think we use mergify here as well, or at least it is nice to use it. Why not fix the config instead? See e.g. https://github.com/crate/crate/commit/48fcd71a9d0cfea9d118987430c6f262694b8b18.

hammerhead:
Looking at [past pull requests](https://github.com/crate/crate-howtos/pulls?q=is%3Apr+is%3Aclosed+label%3Aready-to-merge), the Mergify label was used just once last year, and twice the year before. I can see how Mergify is very useful when dealing with multiple version branches (backports, etc). In simple repository setups like this one, with only a single master branch, what value is Mergify adding?

amotl:
I also think Mergify can be safely removed on this repository. We just don't use it here.

amotl (approved):

seut:
I just wanna note that mergify is not only useful for backports but also, and this is the biggest reason for us at the crate/crate repo, it automates the rebase and merge cycle which is specially useful if multiple PR's on the same repository are opened. This avoids that a developer spent time on rebasing again and again cause some other commits were merged in the mean time.

amotl:
I just updated the Mergify configurations at two other spots. At this place, given the low traffic and criticality about resolving overlapping patches, I think it was safe to remove Mergify here.

Otherwise, on the matter you are outlining, Mergify indeed does a great job.

- https://github.com/crate/crate/pull/13484
- https://github.com/crate/crash/pull/377

PR #319, opened 2023-02-14
Always use `/reference/en/latest` when linking to documentation

hammerhead:
I noticed that there were a few occasions linking to `/reference/en/master` or `/reference/en/4.4` for no specific reason. This updated those occasions to always use `/reference/en/latest`.

## Checklist

 - [X] [CLA](https://crate.io/community/contribute/cla/) is signed


amotl (approved):
Thank you! 👍

PR #322, opened 2023-06-07
Introduce upgrade guidelines

karynzv:
This page aims to define a set of best practices to achieve a smooth upgrade process. Any type of deployment can use these steps.

We want to avoid any specific instructions containing code or version-specific details.

All feedback is welcome!

## Checklist


amotl:
Thank you for providing a good outline on this matter. I would like to add more details on a few items, but maybe we can do it on a later iteration. Because I will have to educate myself about a few aspects first, I am asking for suggestions from @seut and @matriv.

docs/admin/upgrade-general-guidelines.rst
```diff
@@ -0,0 +1,65 @@
...
+
+- Assess impact
+
+- Restore from backup
+
+- Perform data validation
+
+- Evaluate the failed upgrade attempt
+
+- Share insights
```
amotl, line 51: Those sections could be grouped into a "Planning" section.

docs/admin/upgrade-general-guidelines.rst
```diff
@@ -0,0 +1,65 @@
...
+
+
+Next step
+---------
+
+Choose the upgrade strategy that works best for your scenario.
+
+- :ref:`rolling_upgrade` 
+
+- :ref:`full_restart_upgrade`
```
amotl, line 61: After the "Planning" phase, this would be the "Execution" phase. I did not follow up reading the target resources yet in the context of general usability improvements, but what comes to mind here is that we should recommend to check that your cluster is not on fire in any way, before starting with the upgrade procedure on the production database. In this spirit, specific metrics or queries how to determine that, should be explained in detail, including usage guidelines.

docs/admin/general-upgrade-guidelines.rst
```diff
@@ -0,0 +1,73 @@
+.. highlight:: sh
+
+.. _general_upgrade_guidelines:
+
+==========================
+General upgrade guidelines
+==========================
+
+Before kicking off an upgrade, there is a set of best practices to ensure the best outcome. Below you may find the five fundamental steps to prepare for an upgrade.
```
hammerhead, line 9: >  five fundamental steps to prepare for an upgrade

Isn't it just four? 1) Acknowledge breaking changes, 2) Set up a test environment, 3) Back up and plan recovery, 4) Define a rollback plan.

docs/admin/general-upgrade-guidelines.rst
```diff
@@ -0,0 +1,73 @@
...
+   This is not an exhaustive list, so you should consider your organization's specific needs and incorporate any additional steps or considerations that are relevant to your environment.
+
+Upgrade planning
+================
+
+Acknowledge breaking changes
+-----------------------------
+
+Review the `release notes`_ and documentation for the target version to understand any potential impact on existing functionality. 
+In case of a minor version upgrade, you should consider the intermediate versions. For example, when upgrading from 5.1 to 5.3, besides reviewing 5.3 release notes, the recommendation is to check for version 5.2 as well.
```
hammerhead, line 22: > In case of a minor version upgrade, you should consider the intermediate versions

That also applies to major version upgrades (e.g. from 4.x to 5.y), the minor versions in between also need to be checked.

> the recommendation is to check for version 5.2 as well

I would phrase that more strongly, it's not just recommended, it needs to be done. Otherwise, breaking changes will be missed.

docs/admin/general-upgrade-guidelines.rst
```diff
@@ -0,0 +1,73 @@
...
+For the newly written records, you should consider using a mechanism to queue them (e.g. message queue), so these messages can be replayed if needed.
+
+.. WARNING::
+   
+   Before starting the upgrade process, ensure no backup processes are triggered, so disable any scheduled backup.
+
+Define a rollback plan
+-----------------------
+
+The rollback plan may vary depending on the specific infrastructure and upgrade process in use. It is also essential to adapt this outline to your organization's specific needs and incorporate any additional steps or considerations that are relevant to your environment. A set of steps to serve as an example is listed below:
```
hammerhead, line 44: One additional step would also be to identify the rollback process, depending on the version jump. It doesn't necessarily need to be a backup restore. E.g. if upgrading from one patch release to another, and there is no data corruption but only a performance issue, a simple in-place downgrade to the previous patch release is sufficient.
Restoring from a backup would only be needed for major/minor version jumps or in case of data corruption.

docs/admin/general-upgrade-guidelines.rst
```diff
@@ -0,0 +1,73 @@
...
+
+* **Identify the issue** Determine the specific problem that occurred during the upgrade. This could be related to data corruption, performance degradation, application errors, or any other issue that affects the normal functioning of CrateDB.
+
+* **Communicate the situation** Notify all relevant stakeholders, including individuals involved in the upgrade process. Clearly explain the problem and the decision to initiate a rollback.
+
+* **Assess impact** Identify if there are any potential risks to the system's stability, security, or performance.
+
+* **Restore from backup** Restore the recently created backup and replay the latest messages from the message queue to reestablish the CrateDB cluster data.
```
hammerhead, line 52: As per above, here we need to differentiate a bit more I believe, as executing the rollback doesn't necessarily need to be a backup restore.

docs/admin/general-upgrade-guidelines.rst
```diff
@@ -0,0 +1,73 @@
...
+Set up a test environment
+-------------------------
+
+Create a test environment that closely resembles your production environment, including the same CrateDB version, hardware, and network configuration. Populate the test environment with representative data and perform thorough testing to ensure compatibility and functionality, including functional and non-functional testing.
+
+
+Back up and plan recovery
+-------------------------
+
+Perform a cluster-wide backup of your production CrateDB and ensure you have a reliable recovery mechanism in place. Read more in the `snapshots`_ documentation. Once the backup is complete, validate it to ensure its usability.
```
hammerhead, line 33: >  validate it to ensure its usability

Although it's a very valid point, as we figured it's not feasible to actually do so (no built-in validation feature/testing an actual restore takes too long), I guess we can skip that point here.

docs/admin/general-upgrade-guidelines.rst
```diff
@@ -0,0 +1,73 @@
+.. highlight:: sh
+
+.. _general_upgrade_guidelines:
+
+==========================
+General upgrade guidelines
+==========================
+
```
hammerhead, line 8: A table of contents would be nice, as all other pages in the admin section have it as well:

```rst
.. rubric:: Table of contents

.. contents::
   :local:
```

docs/admin/general-upgrade-guidelines.rst
hlcianfagna: If I do an upgrade I feel more confident checking all the release notes, but I think they have it as a policy to only do breaking changes on major releases.

hammerhead (approved):

